Issue #4, opened 2016-06-13
Add initial build scripts


alanshaw:
* [x] start - script to start
* [x] lint+test+coverage - run tests and collect coverage
* [x] build - any js build steps
* [x] minify - any js minification steps

PR #10, opened 2016-06-13
adds semistandard linter #4

Danwhy:
Adds semistandard linter and runs it with the tests.

Also fixes the example test file to comply with semistandard rules.


alanshaw:
👓

PR #14, opened 2016-06-14
Add pre-commit hook

alanshaw:
Runs `npm test` on commit


tobystokes:
Is it best to stop running npm watch while committing?

PR #16, opened 2016-06-14
Scripts

Danwhy:
Adds all currently necessary build scripts and their relevant dependencies


alanshaw:
👓

package.json
```diff
@@ -8,14 +8,37 @@
...
   "scripts": {
-    "test": "npm run lint && istanbul cover tape test/*.js",
-    "lint": "semistandard"
+    "postinstall": "run-s build minify",
+    "start": "node index.js",
+    "test": "run-s test:*",
+    "test:lint": "semistandard",
+    "test:unit": "istanbul cover tape test/*.js",
+    "build": "run-p build:*",
+    "build:js": "browserify ./index.js -o public/bundle.js",
```
alanshaw: I imagine the client code we'll browserify will have a different entry point to the server!

package.json
```diff
@@ -8,14 +8,37 @@
...
+    "build:css": "node-sass ./*.scss public/bundle.css",
+    "minify": "run-p minify:*",
+    "minify:js": "uglifyjs public/bundle.js -o public/bundle.js",
+    "minify:css": "postcss -u cssnano -i public/bundle.css -o public/bundle.css",
+    "watch": "npm-run-all build --parallel watch:*",
+    "watch:js": "watchify ./index.js -o public/bundle.js",
```
alanshaw: Same here, I think we'll want to watch `client.js` (for example) right now...

alanshaw:
Looks good otherwise, if you can add a dummy `client.js` file for browserifying right now that would be awesome :thumbsup:

Danwhy:
No problem

PR #26, opened 2016-06-14
Content negotiation

Danwhy:
Uses hapi-negotiator plugin to deal with differentiating between html and JSONAPI requests. Still to do: need to make sure we comply with these rules: http://jsonapi.org/format/#content-negotiation


alanshaw:
👓

test/content-negotiation-test.js
```diff
@@ -0,0 +1,61 @@
...
+
+  server.inject(badJSONRequest, function (res) {
+    t.equal(res.statusCode, 406, 'JSONAPI request with parameter is not acceptable');
+  });
+
+  server.inject(acceptableJSONRequest, function (res) {
+    t.equal(res.statusCode, 200, 'One JSONAPI without parameters should work');
+    t.end();
+  });
+});
```
alanshaw: I think it's better to split these out into separate tests. Right now this all works because the response is synchronous, but as soon as it becomes async `t.end()` could be called before the other requests have returned.

test/content-negotiation-test.js
```diff
@@ -0,0 +1,61 @@
...
+  };
+
+  var acceptableJSONRequest = {
+    method: 'GET',
+    url: '/',
+    headers: {'Accept': 'application/vnd.api+json; charset=utf-8, application/vnd.api+json'}
+  };
+
+  server.inject(htmlRequest, function (res) {
+    t.equal(res.payload, '<h1>HTML Response</h1>', 'HTML request correct');
```
alanshaw: It won't be possible to test the full response payload payload like this in the future. I suggest we look for HTML we know will be there, like `<!doctype html>` for example.

test/content-negotiation-test.js
```diff
@@ -0,0 +1,61 @@
...
+  var badJSONRequest = {
+    method: 'GET',
+    url: '/',
+    headers: {'Accept': 'application/vnd.api+json; charset=utf-8'}
+  };
+
+  var acceptableJSONRequest = {
+    method: 'GET',
+    url: '/',
+    headers: {'Accept': 'application/vnd.api+json; charset=utf-8, application/vnd.api+json'}
```
alanshaw: I'm not sure I understand what this is testing, splitting tese test cases out into separate tests with an appropriate description will hopefully help!

test/content-negotiation-test.js
```diff
@@ -0,0 +1,61 @@
...
+
+server.route(routes);
+
+server.register(require('hapi-negotiator'), function (err) {
+  if (err) {
+    console.error('Failed to load plugin:', err);
+  }
+});
+
+test('Content Negotiation', function (t) {
```
alanshaw: It's a good idea to always `t.plan` as the first thing you do in the test - this helps prevent false positives when assertions in conditionals or async code you expect to be run are (for whatever reason) not!

test/content-negotiation-test.js
```diff
@@ -0,0 +1,61 @@
...
+  port: process.env.TESTPORT || '8080'
+});
+
+server.route(routes);
+
+server.register(require('hapi-negotiator'), function (err) {
+  if (err) {
+    console.error('Failed to load plugin:', err);
+  }
+});
```
alanshaw: We now need to refactor our `server.js` file to return a function that takes config and a callback as its argument. We can then reuse it here to test the actual server code.

e.g.

**server.js**
```js
module.exports = (config, cb) => {
  const server = new Hapi.Server()

  server.connection({ port: config.port })
  server.route(routes)

  server.register([require('inert'), require('hapi-negotiator')], (err) => {
    if (err) return cb(err)
    cb(null, server)
  })
}
```

I suggest we create a `testWithServer` helper module that will create a new server for us to test against, and we might use it like this:

```js
testWithServer('Should do something', (t, server) => {
  // ...
})
```

For running the server we need a `bin/server.js` which will setup and start a new server for us:

```js
const config = require('../config')
const createServer = require('../server')

createServer(config, (err, server) => {
  if (err) throw err

  server.start(() => {
    if (err) throw err
    console.log(server.info.uri)
  })
})
```
The `npm start` script will need to be changed to call this file instead.

test/content-negotiation-test.js
```diff
@@ -0,0 +1,61 @@
+var test = require('tape');
+var Hapi = require('hapi');
+var server = new Hapi.Server();
+var routes = require('../routes.js');
```
alanshaw: Since we're using node 6 it might be a better idea to express that ee don't intend to re-assign to these variables by declaring them as `const`.

Also, ".js" isn't required in `require` calls.

routes.js
```diff
@@ -4,6 +4,20 @@ module.exports = [
...
+            'application/vnd.api+json': function (req, reply) {
+              reply('"{"response": "JSONAPI"}"').header('content-type', 'application/vnd.api+json');
+            }
+          }
+        }
+      }
+    },
     handler: function (req, reply) {
       reply('OK');
     }
```
alanshaw: Just a note:

If an acceptable media type is not found, hapi-negotiator will return [406 Not Acceptable](https://github.com/felipeleusin/hapi-negotiator/blob/master/lib/index.js#L29) so it makes this handler redundant, but Hapi requires there to be a handler so there needs to be something here :(

alanshaw:
:thumbsup:

PR #28, opened 2016-06-15
Static test

Danwhy:
Tests for the serving of static files. Had to add quotes to the test filenames in package.json to stop tape from occasionally hanging: https://github.com/substack/tape/issues/223


Danwhy:
I'll rewrite these tests using the new server

alanshaw:
Could you also split into multiple tests using `plan` and `end` at the same time?

Danwhy:
Yeah, will do that too

PR #29, opened 2016-06-15
Server refactor

Danwhy:
* Refactors the server to take in config options
* Rewrites server based tests with a helper function that creates a server and calls tape

### DIFF
--- a/bin/server.js
+++ b/bin/server.js
@@ -1,0 +1,15 @@
+const config = require('../config');
+const createServer = require('../server');
+
+createServer(config, (err, server) => {
+  if (err) {
+    throw err;
+  }
+
+  server.start(() => {
+    if (err) {
+      throw err;
+    }
+    console.log(server.info.uri);
+  });
+});
--- a/bin/server.js
+++ b/bin/server.js
@@ -6,7 +6,7 @@ createServer(config, (err, server) => {
     throw err;
   }
 
-  server.start(() => {
+  server.start((err) => {
     if (err) {
       throw err;
     }
--- a/test/helpers/test-with-server.js
+++ b/test/helpers/test-with-server.js
@@ -8,15 +8,8 @@ module.exports = (description, cb) => {
       throw err;
     }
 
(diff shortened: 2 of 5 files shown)


alanshaw:
👓

test/helpers/test-with-server.js
```diff
@@ -0,0 +1,22 @@
...
+const createServer = require('../../server');
+const config = require('../../config');
+
+module.exports = (description, cb) => {
+  createServer(config, (err, server) => {
+    if (err) {
+      throw err;
+    }
+
+    server.start(() => {
```
alanshaw: We don't need to start the server, just create it.

test/helpers/test-with-server.js
```diff
@@ -0,0 +1,22 @@
...
+      if (err) {
+        throw err;
+      }
+
+      tape(description, (t) => {
+        cb(t, server);
+        server.stop();
```
alanshaw: For async operations this will probably stop the server before we've finished testing it.

Since we're probably going to be using `server.inject` in our tests we don't actually need to "start" the server.

test/helpers/test-with-server.js
```diff
@@ -0,0 +1,22 @@
...
+
+module.exports = (description, cb) => {
+  createServer(config, (err, server) => {
+    if (err) {
+      throw err;
+    }
+
+    server.start(() => {
+      if (err) {
+        throw err;
```
alanshaw: This is unreachable as it references `err` in the parent callback.

bin/server.js
```diff
@@ -0,0 +1,15 @@
...
+const createServer = require('../server');
+
+createServer(config, (err, server) => {
+  if (err) {
+    throw err;
+  }
+
+  server.start(() => {
+    if (err) {
+      throw err;
```
alanshaw: Also unreachable - you need to add `err` as a parameter to this callback :)

alanshaw:
LGTM

Issue #32, opened 2016-06-16
Create simple API search route


alanshaw:
* Connect to elastic search
* Submit query
* Retrieve results

PR #36, opened 2016-06-17
Adds elasticsearch dependency and error handler plugin

alanshaw:
refs #32 

* Adds elasticsearch dependency to the app, creates a new client based on .corc config and passes it to `server.js`
* `testWithServer` now creates and passes a noop mock elasticsearch client to the server which you can use to stub/spy/mock on
* Adds sinon dependency for stubbing/spying/mocking
* Adds error handler plugin for Hapi that responds correctly given the request `Accept` header (sends a nice page for text/html and sends an appropriately formatted JSON response for 'application/vnd.api+json')
* Adds tests for the plugin

For this to work you'll need to add elasticsearch config to your `.corc` file e.g.:

```js
{
  "port": 8000,
  "elasticsearch": {
    "apiVersion": "1.7",
    "host": "http://USERNAME:PASSWORD@DOMAIN/public",
    "log": "trace"
  }
}
```

### DIFF
--- a/.corc.template
+++ b/.corc.template
@@ -2,7 +2,6 @@
   "port": 8000,
   "elasticsearch": {
     "apiVersion": "1.7",
-    "host": "",
-    "path": ""
+    "host": ""
   }
 }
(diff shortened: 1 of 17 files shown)


Danwhy:
👓

PR #67, opened 2016-06-21
Json resource

Danwhy:
When a jsonapi request is made to the server, passing in a resource id: 
* Gets the resource from the elasticsearch db
* Separates out its attributes and relationships
* Generates a link to itself on the smgcollections site
* Sends data back as type application/vnd.api+json

Routes to access the pages have changed slightly, you now need to pass an id as a parameter (anything will do for now), I've updated the readme to reflect this.

Todo:
* Use Type mapping module to transform internal/external types
* Separate maker out from lifecycle attribute into a relationship


alanshaw:
👓

README.md
```diff
@@ -50,3 +50,16 @@ The name of the index types isn't always obvious to the public so on the site th
...
+
+### Routes
+
+To access the pages we currently have templates for, use:
+
+* `/`
+* `/search?q={query}`
+* `/object/{id}`
+* `/person/{id}`
+* `/archive/{id}`
```
alanshaw: Could you fix these up so they reflect the API docs: https://github.com/TheScienceMuseum/collectionsonline/wiki/Collections-Online-API#get-objectsidslug

lib/jsonapi-response.js
```diff
@@ -0,0 +1,62 @@
...
+
+  response.data = {};
+  response.data.type = data._type;
+  response.data.id = data._id;
+  response.data.attributes = getAttributes(data._source);
+
+  response.relationships = getRelationships(data._source);
+
+  response.links = {};
+  response.links.self = 'http://www.sciencemuseum.org.uk/' + data._type + '/' + data._id;
```
alanshaw: Config now has a [`rootUrl`](https://github.com/TheScienceMuseum/collectionsonline/blob/master/.corc.template#L3) item which you should be able to pass through here

alanshaw: The type and id should be mapped using the `type-mapping.js` module

lib/jsonapi-response.js
```diff
@@ -0,0 +1,62 @@
...
+    if (relationshipsMap.indexOf(key) > -1) {
+      relationships[key] = [];
+
+      data[key].forEach((el, i) => {
+        let rel = {
+          data: {
+            type: '',
+            id: '',
+            attributes: {}
+          },
```
alanshaw: According to the spec, relationship data should be a "[resource linkage](http://jsonapi.org/format/#document-resource-object-linkage)" object. This effectively means it just has `type` and `id` properties.

The actual linked item should be (deduped) in the top level `included` array http://jsonapi.org/format/#document-compound-documents

lib/jsonapi-response.js
```diff
@@ -0,0 +1,62 @@
...
+          links: {}
+        };
+
+        rel.data.id = el.admin.uid;
+        rel.data.type = el.admin.id.split('-')[0];
+
+        rel.data.attributes.summary_title = el.summary_title;
+
+        rel.links.related = 'http://www.sciencemuseum.org.uk/' + rel.data.type + '/' + rel.data.id;
```
alanshaw: This is the `links.self` link not related.

routes/archive.js
```diff
@@ -18,7 +19,13 @@ module.exports = () => ({
             reply.view('archive', Object.assign(exampleData, data));
           },
           'application/vnd.api+json' (req, reply) {
-            reply('"{"response": "JSONAPI"}"').header('content-type', 'application/vnd.api+json');
+            elastic.get({index: 'smg', type: 'archive', id: req.params.id}, (err, result) => {
+              if (err) {
+                return reply(err);
+              }
```
alanshaw: We need a test after this (in all item routes) for no result:

```js
if (!result) return reply(Boom.notFound('Document not found'))
```

test/jsonapi-response-builder.test.js
```diff
@@ -0,0 +1,13 @@
...
+const buildJSONResponse = require('../lib/jsonapi-response.js');
+const exampleGETResponse = require('./fixtures/example-get-response.json');
+
+test('Building the JSON Response', function (t) {
+  t.plan(3);
+  const JSONAPIResponse = JSON.parse(buildJSONResponse(exampleGETResponse));
+
+  t.ok(JSONAPIResponse.data, 'Response contains data field' + JSONAPIResponse.data);
+  t.ok(JSONAPIResponse.relationships, 'Response contains relationships field');
+  t.ok(JSONAPIResponse.links, 'Response contains links field');
```
alanshaw: Might be nice to verify some properties of these fields are as expected!

alanshaw:
Looks good, just some minor tweaks.

I'd explicitly split the creators for results by type so it will be interesting to see if `lib/jsonapi-response.js` will work out. It's possible I over complicated this for search results unnecessarily.

Danwhy:
Ok, I'll get it fixed up

alanshaw:
Apologies I missed your todo about type mapping 🚀

Danwhy:
This should be good apart from the type mapping now. I can do that in a separate PR though

routes/object.js
```diff
@@ -24,6 +25,10 @@ module.exports = ({ elastic }) => ({
                 return reply(err);
               }
 
+              if (!result) {
+                return reply(Boom.notFound('Document not found'));
```
alanshaw: "Object not found"

routes/person.js
```diff
@@ -24,6 +25,10 @@ module.exports = ({ elastic }) => ({
                 return reply(err);
               }
 
+              if (!result) {
+                return reply(Boom.notFound('Document not found'));
```
alanshaw: "Person not found"

lib/jsonapi-response.js
```diff
@@ -0,0 +1,102 @@
+const attributesMap = require('../fixtures/attributes');
+const relationshipsMap = require('../fixtures/relationships');
+const config = require('../config');
```
alanshaw: I'm generally more included to pass config in to modules/functions as a parameter. The same going for any stateful dependencies (like elasticsearch client in our case).

Requiring config in a module actually creates a dependency on whatever config module we're using to configure the app (in our case `rc`) and makes it a little more difficult to test in isolation.

In this case it's not a massive problem, we can just require the config in the test to grab the `rootUrl` so we can make our assertions. Imagine for example if the module used some secrets to connect to stripe or something - in an integration test you probably don't want whatever credentials you have in config for development, you probably want to pass it config for connecting to the test Stripe server and using test credentials.

Put it this way, if you were to extract this module and publish it to npm, how would you allow consumers to set the root URL - you probably wouldn't enforce use of a particular config module and would probably instead pass it in as a parameter to the function.

https://github.com/TheScienceMuseum/collectionsonline/blob/master/lib/transforms/search-results-to-jsonapi.js#L4

lib/jsonapi-response.js
```diff
@@ -0,0 +1,102 @@
...
+};
+
+function getAttributes (data, dataType) {
+  const attributes = {};
+
+  Object.keys(data).forEach((key) => {
+    if (attributesMap.indexOf(key) > -1) {
+      attributes[key] = data[key];
+    }
+  });
```
alanshaw: It would be quicker to loop over `attributesMap` and add it to `attributes` if it exists on `data` - would that work or am I missing something?

lib/jsonapi-response.js
```diff
@@ -0,0 +1,102 @@
...
+  response.data.attributes = getAttributes(data._source);
+
+  response.relationships = getRelationships(data._source);
+
+  response.included = getIncluded(data._source, response);
+
+  response.links = {};
+  response.links.self = config.rootUrl + '/' + data._type + '/' + data._id;
+
+  return JSON.stringify(response);
```
alanshaw: You shouldn't need to stringify this - Hapi will JSON stringify replys by default

lib/jsonapi-response.js
```diff
@@ -0,0 +1,102 @@
...
+  });
+
+  return attributes;
+}
+
+function getRelationships (data) {
+  const relationships = {};
+
+  Object.keys(data).forEach((key) => {
+    if (relationshipsMap.indexOf(key) > -1) {
```
alanshaw: Again, loop over `relationshipsMap` instead?

lib/jsonapi-response.js
```diff
@@ -0,0 +1,102 @@
...
+
+        rel.id = el.admin.uid;
+        rel.type = el.admin.id.split('-')[0];
+
+        relationships[key].data.push(rel);
+      });
+
+      if (relationships[key].data.length === 1) {
+        relationships[key].data = relationships[key].data[0];
+      }
```
alanshaw: Is there something in the spec that mandates this? I think it's easier to reason about if `relationships.data` is always an array, even if there's only one.

lib/jsonapi-response.js
```diff
@@ -0,0 +1,102 @@
...
+
+      if (relationships[key].length === 1) {
+        relationships[key] = relationships[key][0];
+      }
```
alanshaw: Same as https://github.com/TheScienceMuseum/collectionsonline/pull/67/files#r67857651

lib/jsonapi-response.js
```diff
@@ -0,0 +1,102 @@
+const attributesMap = require('../fixtures/attributes');
+const relationshipsMap = require('../fixtures/relationships');
```
alanshaw: I think these could be named better - "map" suggests they're key => value, whereas they're arrays. Not that I'm suggesting renaming Map => Array since I don't think it's right to name variables after the type of the value they hold!

What about `allAttributes` or `allRelationships`?

lib/jsonapi-response.js
```diff
@@ -0,0 +1,102 @@
...
+        relationships[key] = relationships[key][0];
+      }
+    }
+  });
+
+  return relationships;
+}
+
+function getIncluded (data, response) {
+  if (response.relationships) {
```
alanshaw: This is always going to be truthy since we return `{}` from `getRelationships` even if there are none

lib/jsonapi-response.js
Danwhy: For this one, it should always be just one object, so I didn't think an array made sense

alanshaw:
Awesome job, thanks! :thumbsup:

PR #78, opened 2016-06-21
Makers

Danwhy:
Maker and place of creation aren't direct attributes of the data, they're nested inside the lifecycle attribute, so this is needed to pull them out and add them as relationships


alanshaw:
👓

lib/jsonapi-response.js
```diff
@@ -45,46 +45,68 @@ function getRelationships (data) {
...
-        rel.type = TypeMapping.toExternal(el.admin.id.split('-')[0]);
-
-        relationships[key].data.push(rel);
+        if (el.admin) {
+          rel.id = TypeMapping.toExternal(el.admin.uid);
+          rel.type = TypeMapping.toExternal(el.admin.id.split('-')[0]);
+          relationships[key].data.push(rel);
+        } else {
+          addAsAttribute(el, key, response);
+        }
```
alanshaw: I see what you mean here, but I'd be inclined to ignore this relationship entirely if it has no ID to keep to the spec.

lib/jsonapi-response.js
```diff
@@ -45,46 +45,68 @@ function getRelationships (data) {
...
       });
 
       if (relationships[key].length === 1) {
         relationships[key] = relationships[key][0];
       }
```
alanshaw: Can `relationships[key]` ever be an array? I don't think we need these lines!

lib/jsonapi-response.js
```diff
@@ -45,46 +45,68 @@ function getRelationships (data) {
...
+
+      if (!relationships[key].data.length) {
+        delete relationships[key];
+      }
     }
   });
 
-  return relationships;
+  const creationDetails = getCreationDetails(data, config);
```
alanshaw: In this file we seem to be augmenting objects passed to the functions, but this one is different and returns an object to be merged in.

In general we should try to create functions that operate on parameters and return a value. Augmenting objects couples the function to the data passed in and will cause it to make assumptions about its structure that you can't guarantee or always guard against in every case. Augmenting is what's commonly known as a side effect which should be avoided as much as possible to make your code more robust and testable.

It's up to you how you want to handle this, it might be an idea to just be consistent and pass the relationships in to be augmented.

As a comparison, the search results has a `createRelationships` function that returns an object with `relationships` and `included` properties - https://github.com/TheScienceMuseum/collectionsonline/blob/master/lib/transforms/search-results-to-jsonapi.js#L68

lib/jsonapi-response.js
```diff
@@ -10,7 +10,7 @@ module.exports = (data, config) => {
   response.data.id = TypeMapping.toExternal(data._id);
   response.data.attributes = getAttributes(data._source);
 
-  response.relationships = getRelationships(data._source);
+  response.relationships = getRelationships(data._source, response, config);
```
alanshaw: You're already mentioned this but relationships should be on the `data` property.

test/jsonapi-response-builder.test.js
```diff
@@ -1,11 +1,13 @@
...
 const config = require('../config');
 const buildJSONResponse = require('../lib/jsonapi-response.js');
-const exampleGETResponse = require('./fixtures/example-get-response.json');
+const exampleObjResponse = require('./fixtures/example-get-response-object.json');
+const exampleDocResponse = require('./fixtures/example-get-response-document.json');
 
-test('Building the JSON Response', function (t) {
-  t.plan(11);
-  const JSONAPIResponse = buildJSONResponse(exampleGETResponse, config);
+test('Building the JSON Response for an object', function (t) {
```
alanshaw: I'd recommend naming your tests in terms of what they "should" or "should not" do. It encourages you to narrow down what you're testing, be a bit more descriptive about what the test is testing and split apart big tests like this into smaller chunks that test a specific aspects.

For instance, I'd be tempted to split this into multiple tests to test that the builder adds the fields required for a jsonapi response -

* "Should create data field" - make sure data is present and has required attributes
* "Should create relationships within data" - make sure terms, maker and places relations are added
* "Should create included field" - make sure included relationships are in the response

I'd then suggest another file for documents, and people that does the same.

Perhaps think of the filename as the "suite", `test()` as the test and `t` as assertions. Right now you seem to have the `test()` as the suite and your assertions (`t`) as the tests.

lib/jsonapi-response.js
```diff
@@ -96,3 +106,40 @@ function getIncluded (data, response, config) {
     return included;
   }
 }
+
+function getCreationDetails (data, config) {
```
alanshaw: `config` unused in this function

lib/jsonapi-response.js
```diff
@@ -32,7 +26,7 @@ function getAttributes (data) {
   return attributes;
 }
 
-function getRelationships (data) {
+function getRelationships (data, response, config) {
```
alanshaw: `response` and `config` not used (config only used to be passed to `getCreationDetails` but it is unused here too)

lib/jsonapi-response.js
```diff
@@ -45,46 +39,62 @@ function getRelationships (data) {
...
                 resource.attributes[key] = el[key];
               }
             });
-          });
-        }
+          }
+        });
 
         resource.links = {};
         resource.links.self = config.rootUrl + '/' + resource.type + '/' + resource.id;
```
alanshaw: Really minor, but probably more appropriate to use a template string here...

alanshaw:
@Danwhy looks good, a few final points to clean up but then good to merge :thumbsup:

Danwhy:
@alanshaw should be good to go

Issue #83, opened 2016-06-22
Sprites/Icons not displaying in Safari (since move to this repo)


jamieu:
### Sprites not displaying in Safari (9.1.1)

Although:

- Works fine in Chrome / Firefox
- The sgv file seems to load fine (/assets/icons/symbol/svg/sprite.symbol.svg#grid) icons just aren't displaying.
- The icons are displaying in Safari in the old codesign repo

**Page in Safari in this repo (icons/sprites not working):**

<img width="1244" alt="safari-no-icons-in-new-co-repo" src="https://cloud.githubusercontent.com/assets/91365/16264168/dcc1ac7a-386f-11e6-9a27-f703b3ca880d.png">

**Same page in old repo in Safari in (icons work):**

<img width="1191" alt="safari-icons-old-codesign-repo" src="https://cloud.githubusercontent.com/assets/91365/16264300/af1dd6da-3870-11e6-9032-931ca18765cf.png">

PR #84, opened 2016-06-22
Asset fix

Danwhy:
Renames assets to correct external types so they correctly show.

Also deletes two empty example files we don't use


alanshaw:
Is this to fix #83?

Danwhy:
No, this is a separate issue where the svgs weren't displaying anymore on any browser due to the type name changing from 'person' to 'people' and 'object' to 'objects'

alanshaw:
LGTM

PR #88, opened 2016-06-22
Resource template (People)

### LINKED ISSUE #43: Wire up person result page
* Use existing code to fetch resource from elasticsearch and construct JSONAPI response
* Transform JSONAPI response to appropriate format for Handlebars template
* Serve and display data on page

Depends on #47

Danwhy:
Transforms JSONapi data into data for the person template and renders it. A lot of this code will be relevant to the objects/documents too.

resolves #43


alanshaw:
👓

lib/transforms/json-to-html-data.js
```diff
@@ -0,0 +1,143 @@
...
+function getOccupation (data) {
+  try {
+    if (data.attributes.occupation) {
+      return {key: 'Occupation', value: data.attributes.occupation};
+    } else {
+      throw new Error('No Occupation');
+    }
+  } catch (err) {
+    return null;
+  }
```
alanshaw: What's the reason for using try/catch like this? We should be using try/catch for exceptions, meaning simply that I don't think we should regard properties that legitimately may or may not exist on an object as an exception.

alanshaw: If you want to avoid checking if each property exists on an object, it's probably an idea to use a module that'll do that for you.

routes/person.js
```diff
@@ -14,11 +13,23 @@ module.exports = ({ elastic, config }) => ({
...
             };
 
-            reply.view('person', Object.assign(exampleData, data));
+            elastic.get({index: 'smg', type: 'agent', id: TypeMapping.toInternal(request.params.id)}, (err, result) => {
+              if (err) {
+                return reply(err);
+              }
+
+              if (!result) {
+                return reply.view('noresults');
```
alanshaw: Need to set the status code here to 404

templates/helpers/format-description.js
```diff
@@ -0,0 +1,28 @@
+const Handlebars = require('handlebars');
+
+module.exports = (description) => {
+  const splitDescription = description.split('\n');
+  let initialDescription;
+  let moreDescription;
```
alanshaw: I think we can't use `let` in Node.js 4

templates/helpers/format-description.js
```diff
@@ -0,0 +1,28 @@
...
+          ${formattedDescription.slice(15).join('')}
+        </div>
+      </div>`;
+
+    return new Handlebars.SafeString(
+      `${initialDescription.join('')}${moreDescription}`
+    );
+  } else {
+    return new Handlebars.SafeString(formattedDescription.join(''));
+  }
```
alanshaw: This all needs to be done before render time in `json-to-html-data.js`.

The description needs to be split into two fields, the handlebars template needs to then detect the presence of the `moreDescription` field and show the HTML for the "More..." expander if necessary.

It keeps HTML in the templates and all the logic for data transformations in `json-to-html-data.js`.

templates/helpers/related.js
```diff
@@ -0,0 +1,17 @@
...
+  if (relation.attributes && relation.attributes.role) {
+    role = `(${relation.attributes.role})`;
+  } else {
+    role = '';
+  }
+
+  return new Handlebars.SafeString(
+    '<li><a href="' + url + '">' + name + '</a> ' + role + '</li>'
+  );
+};
```
alanshaw: The HTML should be a handlebars template and the processing beforehand should be done by `json-to-html-data.js`

test/transforms/json-to-html-org.test.js
```diff
@@ -0,0 +1,17 @@
+const test = require('tape');
+const config = require('../../config');
+const buildJSONResponse = require('../../lib/jsonapi-response');
+const JSONAPIResponse = buildJSONResponse(require('../fixtures/example-get-response-organisation.json'), config);
+const buildHTMLData = require('../../lib/transforms/json-to-html-data');
+const HTMLData = buildHTMLData(JSONAPIResponse);
```
alanshaw: This should be done within the test - since one thing you can test for is whether it throws!

alanshaw:
Getting an error for this person:
screen shot 2016-06-23 at 16 14 20

Please add a test.

Danwhy:
@alanshaw That error is actually from the json response. Do you mind if I fix that in a separate PR? Other than that, this should hopefully be ok now

Danwhy:
That error is now fixed too

lib/helpers/property-exists.js
```diff
@@ -0,0 +1,11 @@
...
+  var args = Array.prototype.slice.call(arguments, 1);
+
+  for (var i = 0; i < args.length; i++) {
+    if (!obj || !obj.hasOwnProperty(args[i])) {
+      return false;
+    }
+    obj = obj[args[i]];
+  }
+  return true;
+};
```
alanshaw: Would this be more useful actually extracting the property from the object and returning it, or returning `null` if it doesn't exist. I suspect that the majority of the time you actually want to retrieve the value and use it, rather than just checking if it exists.

alanshaw: How about dot notation for the properties? i.e.

```js
propertyExists({}, 'foo.bar.0.baz')
```

alanshaw: Minor, but I don't think this needs to go in a helpers folder - it's essentially what the `lib` folder is!

lib/transforms/json-to-html-data.js
```diff
@@ -0,0 +1,132 @@
...
+}
+
+function getBorn (data) {
+  if (propertyExists(data, 'attributes', 'lifecycle', 'birth', '0', 'location', 'name', '0', 'value')) {
+    return {key: 'Born', value: data.attributes.lifecycle.birth[0].location.name[0].value};
+  } else {
+    return null;
+  }
```
alanshaw: If `propertyExists` returned the value this function could be as simple as:

```js
const value = propertyExists(data, 'attributes', 'lifecycle.birth.0.location.name.0.value')
return value ? { key: 'Born', value } : null
```

routes/person.js
```diff
@@ -14,11 +13,32 @@ module.exports = ({ elastic, config }) => ({
...
+                if (err.status === 404) {
+                  const error = {
+                    output: {
+                      payload: {
+                        status: 404,
+                        error: 'Not found',
+                        message: 'The requested person could not be found'
+                      }
+                    }
+                  };
```
alanshaw: Why not `Boom.notFound('The requested person could not be found')`?

routes/person.js
```diff
@@ -14,11 +13,32 @@ module.exports = ({ elastic, config }) => ({
...
+
+                  return reply.view('person', {error}).code(404);
```
alanshaw: No need to reply with a view here - the error handler plugin should handle the error and display an error page, simply `return reply(Boom.notFound())`

templates/partials/records/record-description.html
```diff
@@ -0,0 +1,21 @@
...
+      <div class="details__content">
+        {{#each description.moreDescription}}
+          <p>
+            {{this}}
+          </p>
+        {{/each}}
+      </div>
+    </div>
+  {{/if}}
+{{/if}}
```
alanshaw: :thumbsup: nice

templates/partials/records/record-details.html
```diff
@@ -7,7 +7,7 @@
         <dl class="record-details__dl">
           <dt>{{ this.key }}:</dt>
           <dd>
-            {{#if link}}<a href="#" title="[link to filtered search results]" class="searchlink">{{/if}}
+            {{#if link}}<a href={{link}} title="[link to filtered search results]" class="searchlink">{{/if}}
```
alanshaw: Add double quotes around this just in case it contains a space character

templates/partials/records/record-related-organisations.html
```diff
@@ -1,7 +1,16 @@
...
-          <li><a href="/organisation.html">British Rail</a> (commissioned)</li>
-        </ul>
-      </div>
+<div class="panel panel__text">
+  <h4>Related organisations</h4>
+  <ul class="related-organisation">
+    {{#each related.organisations}}
+    <li>
+      {{#if this.links}}
+        <a href={{this.links.self}}>
```
alanshaw: Quotes here please too

test/jsonapi-response-builder-person.test.js
```diff
@@ -1,21 +1,45 @@
...
   t.plan(5);
   t.ok(JSONAPIResponse.data, 'Response should contain data field');
   t.ok(JSONAPIResponse.data.attributes, 'Data field should contain attributes field');
-  t.notOk(JSONAPIResponse.data.relationships, 'Data field should not contain relationships field');
+  t.ok(JSONAPIResponse.data.relationships, 'Data field should contain relationships field');
   t.ok(JSONAPIResponse.data.links, 'Data field should contain links field');
-  t.notOk(JSONAPIResponse.included, 'Response should not contain included field');
+  t.ok(JSONAPIResponse.included, 'Response should contain included field');
   t.end();
 });
```
alanshaw: Tests that depend on the results of other tests are less portable and maintainable. There's no need to optimise tests, call `buildJSONResponse` in each one.

test/transforms/json-to-html-org.test.js
```diff
@@ -0,0 +1,25 @@
...
+test('Data fields should have correct values', function (t) {
+  t.plan(6);
+  t.equal(HTMLData.title, 'Royal Astronomical Society', 'title should be correct');
+  t.equal(HTMLData.type, 'people', 'type should be correct');
+  t.deepEqual(HTMLData.fact, [], 'facts should be empty');
+  t.equal(HTMLData.related.people.length, 1, 'Data should contain one related person');
+  t.notOk(HTMLData.description, 'Data should contain no description');
+  t.equal(HTMLData.details[0].key, 'Website', 'details should contain a website');
+  t.end();
+});
```
alanshaw: Again, please build the response in both tests

test/transforms/json-to-html.test.js
```diff
@@ -0,0 +1,25 @@
...
+test('Data should have correct fields', function (t) {
+  t.plan(6);
+  t.ok(HTMLData.title, 'Data should contain title');
+  t.ok(HTMLData.type, 'Data should contain type');
+  t.ok(HTMLData.fact, 'Data field should contain facts');
+  t.ok(HTMLData.related, 'Data should contain related');
+  t.ok(HTMLData.description, 'Data should contain description');
+  t.ok(HTMLData.details, 'Data should contain details');
+  t.end();
+});
```
alanshaw: Again, please build the response in both tests

templates/pages/person.html
```diff
@@ -1,65 +1,44 @@
 {{> global/global-header}}
 {{> global/search-main reduced="reduced"}}
 
-<div class="page" role="main">
+{{#if error}}
+  {{> global/error}}
+{{else}}
```
alanshaw: This doesn't need to be here, and complicates the template - let the error plugin handle 404 errors and display a 404 page.

templates/partials/global/error.html
```diff
@@ -0,0 +1,7 @@
+<h1>{{error.output.payload.status}} {{error.output.payload.error}}</h1>
+{{#if error.output.payload.message}}
+  <p>{{error.output.payload.message}}</p>
+{{/if}}
+{{#if isDevelopment}}
+  <pre>{{error.stack}}</pre>
+{{/if}}
```
alanshaw: No need to pull this out into a partial right now - only the error page needs to use it

lib/property-exists.js
```diff
@@ -0,0 +1,13 @@
...
+  var property = path.reduce((prev, curr) => {
+    if (prev[curr]) {
+      return prev[curr];
+    } else {
+      return false;
+    }
+  }, obj);
+
+  return property;
+};
```
alanshaw: I'd be tempted to rename this now since it's no longer doing an existence check, but one for another PR perhaps :)

lib/property-exists.js
```diff
@@ -0,0 +1,13 @@
+module.exports = function checkNestedProperty (obj, pathString) {
+  var path = pathString.split('.');
+
+  var property = path.reduce((prev, curr) => {
+    if (prev[curr]) {
+      return prev[curr];
+    } else {
+      return false;
+    }
+  }, obj);
```
alanshaw: In this case it's better to use a for loop to terminate early, as you have it now we have to traverse the whole `pathString`, and we might end up checking for `(false)['x']`, which could result in a false positive if, for instance the property name was 'constructor'

lib/transforms/json-to-html-data.js
```diff
@@ -0,0 +1,111 @@
...
+
+function getNationality (data) {
+  const value = propertyExists(data, 'attributes.nationality');
+  return value ? {key: 'Nationality', value: value} : null;
+}
+
+function getBorn (data) {
+  const value = propertyExists(data, 'attributes.lifecycle.birth.0.location.name.0.value');
+  return value ? {key: 'Born', value: value} : null;
+}
```
alanshaw: Sweet :thumbsup:

alanshaw:
:thumbsup:

Issue #48, opened 2016-06-20
Integrate page.js


alanshaw:
* Allow client to navigate between pages without round trip to server
* Use JSONAPI to retrieve data for single result page
* Use JSONAPI to retrieve results for subsequent searches

Depends on #40 #41 #42 #43

iteles:
Concluded in sprint 2 👍

PR #94, opened 2016-06-24
Remove bower dependencies

alanshaw:
Instead use npm dependencies and browserify to require them into a single file

No strictly required for #48 but it needed to be done and this is a good time/place ;P


Danwhy:
👓

Danwhy:
👍

PR #99, opened 2016-06-24
Runs autoprefixer as part of css build

### LINKED ISSUE #96: CSS should be auto-prefixed
For required browser compatibility, css needs autoprefixer to add vendor prefixes.

(If we are already using post-css to minify, then suggest we use autoprefixer as a post-css plugin, rather than the old standalone version.)

tobystokes:
fixes #96


alanshaw:
👓

package.json
```diff
@@ -54,7 +54,7 @@
     "test:unit": "istanbul cover tape 'test/**/*.test.js'",
     "build": "run-p build:*",
     "build:js": "browserify client/main.js -o public/bundle.js",
-    "build:css": "node-sass client/styles/main.scss public/bundle.css --include-path node_modules/foundation-sites/scss --include-path node_modules/slick-carousel/slick",
+    "build:css": "node-sass client/styles/main.scss public/bundle.css --include-path node_modules/foundation-sites/scss --include-path node_modules/slick-carousel/slick  && postcss --u autoprefixer -b 'last 2 versions, ie >= 9' -i public/bundle.css -o public/bundle.css",
```
alanshaw: I think it'll be best to split this into two scripts, leaving "build:css" to run them, something like:

```json
{
  "build:css": "run-s build:css:*",
  "build:css:sass": "...",
  "build:css:prefix": "...",
}
```

alanshaw: I think this because it makes it easier to read, and also allows us to run that script in isolation of the others.

tobystokes: gotcha, much prettier.

Issue #105, opened 2016-06-27
display which system people came from


SimonLab:
There are some duplicates people coming from two different system. The goal is to merge the duplicates. 

Short term start to this goal:
- [x] Display a tag "Mimsy" (sgmc) or "AdLib" (smga) on the result people page. The tag information is saved on the "source" property of the people document. The id also contains the information for "smgc-people-139208"

PR #114, opened 2016-06-27
display the system of a person

SimonLab:
see  #105
system


nelsonic:
@SimonLab looks good. 👍 
We should work on getting coverage up to 100% in the coming weeks, let me know if you need any input. 😉

Issue #115, opened 2016-06-28
Display number of result per categories


SimonLab:
![aggregationnumber](https://cloud.githubusercontent.com/assets/6057298/16411093/4d531fa8-3d1d-11e6-83b6-5fbada51c26b.png)

To get these numbers I think the best way is to use aggregation on the "type.base" value of the data. We will need to do a mapping between the type of the database (agent, agents, documents,...) to the categories of the website (people, documents, objects). We also need to update the values when the ajax requests are used!

SimonLab:
This feature is implemented with #118

PR #118, opened 2016-06-28
Add number of result for the categories

SimonLab:
see #115 

@nelsonic do you mind to have a look this PR, thanks.
@Danwhy please add your comments too!

### DIFF
--- a/client/routes/search.js
+++ b/client/routes/search.js
@@ -16,6 +16,7 @@ module.exports = function (page) {
       window.location = ctx.path;
     }
 
+    var searchnav = document.getElementById('searchnav');
     var searchBoxEl = document.getElementById('searchbox');
     var searchResultsEl = document.getElementById('searchresults');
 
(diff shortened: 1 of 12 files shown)


nelsonic:
@SimonLab please assign when ready for review. 👍

client/routes/search.js
```diff
@@ -43,6 +44,7 @@ module.exports = function (page) {
         .then(function (json) {
```
nelsonic: which `fetch` are we using here?
> https://github.com/TheScienceMuseum/collectionsonline/blob/2ada8c6ceca133b7e50d3ee0c835de9b62bad6d0/client/routes/search.js#L3

nelsonic:
👀

nelsonic:
@SimonLab tests & code look good. 👍

client/routes/search.js
SimonLab: We are using fetch-ponyfill to determine if the browser of the user support the [fetch standard](https://fetch.spec.whatwg.org/). If the fetch is supported we use the fetch from the browser otherwise we use the fetch-ponyfill implementation.

Issue #120, opened 2016-06-29
Missing values on object page


jamieu:
Linked to #106 
Duchess of Hamilton
Accession No. #1976-7000
/objects/smgc-objects-205752

Note: category, materials, terms, taxonomy terms, makers, locations, places etc. should all be linkable (back to the search resulst page wi the relevent fiter set)

1) Makers (top of page)

We should list all of these at the top of the page on seperate lines, prefixed by their relationship/role. 
(prior to launch we will need to supress all but a defined/limited set of relationships (maker, designer, inventor etc; which will be supplied), but for now it would be useful to allow testers to see all the values)

``````
"maker": [{
    "@link": {
        "role": [{
            "value": "designer"
        }],
        "type": "reference"
    },
    "admin": {
        "id": "agent-1015",
        "uid": "smgc-agent-1015"
    },
    "summary_title": "Stanier, Sir William Arthur"
}, {
    "@link": {
        "role": [{
            "value": "maker"
        }],
        "type": "reference"
    },
    "admin": {
        "id": "agent-3891",
        "uid": "smgc-agent-3891"
    },
    "summary_title": "London, Midland & Scottish Railway"
}],
``````
2) CREDIT (tombstone block)

credit_line

``````
"legal": {
    "credit_line": "Butlins",
    "legal_status": "permanent collection"
},

``````

3) MEASUREMENTS (tombstone block)

component.measurements[].display

Need to display multiple measurements on seperate lines ie.

length over buffers: 22511 mm,
driving wheel diameter: 2057 mm

``````
    "component": [{
        "measurements": {
            "dimensions": [{
                "dimension": "depth",
                "units": "in",
                "value": "886 1/4"
            }, {
                "dimension": "depth",
                "units": "mm",
                "value": "22511"
            }],
            "display": "length over buffers: 22511 mm,"
        },
        "name": "length over buffers"
    },
    {
        "measurements": {
            "dimensions": [{
                "dimension": "diameter",
                "units": "in",
                "value": "81"
            }, {
                "dimension": "diameter",
                "units": "mm",
                "value": "2057"
            }],
            "display": "driving wheel diameter: 2057 mm,"
        },
        "name": "driving wheel diameter"
    }
``````

4) Type (tombstone block)

``````
"name": [
  {
    "primary": true,
    "type": "object name",
    "value": "steam locomotive"
  }
]
``````

5) Taxonomy (tombstone block)

[...]

Issue #106, opened 2016-06-27
Visual fixes for resource pages


Danwhy:
* [x] - Which attributes to use as facts at the top of the page?
* [x] - Which attributes to use as details at the bottom of the page?
* [ ] - What to display when there are no details/facts/description?

jamieu:
See Missing values on object page #120

Danwhy:
* [ ] Makers
* [ ] Credit
* [ ] Measurements
* [ ] Type
* [ ] Taxonomy

jamieu:
Toby made a **very basic** design for a 'no result' page, not sure if it got copied over from the codesign repo.
/noresults.html

* probably needs some some additional work / revisiting

jamieu:
We've just been discussing the archive / document icons and think we'd prefer to use the 'graphical text' as per the objects.

We should use the 'extent' field (on documents) for this text.

btw. just to clarify:

- archive = the top level 'fonds' records (this is itself a 'document' record, we're just treating it as a special one)
- document = everything else (which can include series, sub-series, folders and the records that actually represent a single document). 

note 1: _we may in future represent the folders & series that sit between the top level 'fonds' record and the leaf node records/documents that sit at the bottom of the tree slightly differently. But for now we're treating everything bar the top-level record as a 'document' with a document icon (even if they in represent a record that could have multiple scanned documents attached to it or sub-folders)._

note 2: _I am asking Toby to design us a new top level icon to represent the 'fonds' or 'archive' record #125 but hopefully this will just be a case of switching the 'archive' icon in the CSS_

Jamie

tobystokes:
Yes agree that may look better. Ideally, given that there aren't that many top level archives, if we could assign them a picture to represent, that would be worthwhile.

jamieu:
@tobystokes yes that's what we're gonna try and do, there a small enough set of them that even Emily and I could pick assign sensible images for them all.

jamieu:
A couple more changes, I have [updated the wiki to reflect the additions](https://github.com/TheScienceMuseum/collectionsonline/wiki/Values-to-display-on-HTML--pages#peopleorganisation-page), but changes / deletions below:

## Object

**Changes:**
- Change the title to use `title.value` rather than `summary_title`

## Person/Organisation template

**Remove:**
- taxonomy

**Comment out:**
[...]

PR #123, opened 2016-06-30
Visual resource

Danwhy:
refs #106 and #120 

Gets relevant data for facts/details section of the individual object pages

Facts: 
screen shot 2016-07-01 at 09 50 13

Details:
screen shot 2016-07-01 at 09 50 30

### DIFF
--- a/lib/transforms/json-to-html-data.js
+++ b/lib/transforms/json-to-html-data.js
@@ -7,7 +7,7 @@ module.exports = (resource) => {
   const related = getRelated(resource);
   const system = getSystem(resource.data);
   const description = getDescription(resource.data);
-  const details = getDetails(resource.data);
+  const details = getDetails(resource);
   const displayLocation = getDisplayLocation(resource.data);
 
   return {title, type, fact, related, description, details, displayLocation, system};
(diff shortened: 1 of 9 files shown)


iteles:
@Danwhy Could we get a screenshot of how you expect this to display for completion please? :blush:

lib/transforms/json-to-html-data.js
```diff
@@ -212,15 +219,13 @@ function getMaterials (data) {
...
-  const value =
-    getNestedProperty(data, 'attributes.component.0.measurements.display') ||
-    getNestedProperty(data, 'attributes.measurements.dimensions.0.value');
-  return value ? {key: 'Measurements', value: value} : null;
-}
-
-function getStatus (data) {
-  const value = getNestedProperty(data, 'attributes.legal.legal_status');
-  return value ? {key: 'Status', value: value} : null;
+  const value = [];
```
nelsonic: if I was being picky, I'd suggest calling this variable `values` and not making it a `const` given that it's being pushed to...

lib/transforms/json-to-html-data.js
```diff
@@ -284,6 +308,30 @@ function getNotes (data) {
...
+      if (el.primary) {
+        value = el.value || null;
+      }
+    });
+  }
+  return value ? {key: 'type', value: value} : null;
+}
+
+function getTaxonomy (resource) {
+  const taxonomy = [];
```
nelsonic: same here is this really a `const` ?
https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Statements/const

nelsonic:
I know this PR isn't assigned to me, but I'm trying to stay up-to-date. code looks good.
Would make a couple of _minor_ adjustments, but looking good. 😉

Danwhy:
@nelsonic :+1: I assigned it to Simon so he could have a look first, then once he's good with it he'll assign it to you to check and merge

lib/transforms/json-to-html-data.js
```diff
@@ -29,10 +29,10 @@ function getFacts (resource) {
   const born = getBorn(resource.data);
   const birth = getBirth(resource.data);
   const death = getDeath(resource.data);
-  const maker = getMaker(resource);
+  const makers = getMakers(resource);
   const made = getMade(resource);
 
-  return [occupation, nationality, born, birth, death, maker, made].filter(el => el);
+  return [occupation, nationality, born, birth, death, made].concat(makers).filter(el => el);
```
SimonLab: I like to use filter(Boolean)  but it's just a preference detail

test/fixtures/elastic-responses/example-get-response-object2.json
```diff
@@ -0,0 +1,443 @@
...
+        "type": "display title",
+        "value": "LMS(R) steam locomotive with tender 'Duchess of Hamilton' 4-6-2 Coronation class, No 46229, 1938"
+      }
+    ],
+    "type": {
+      "base": "object",
+      "record_type": "WHOLE_PART"
+    }
+  }
+}
```
SimonLab: I think we should find a way to use the real data instead of creating manually the fixtures. If the data structure change we need to change all the fixtures manually.

SimonLab:
I just add two notes, the code looks good for me, @nelsonic what do you think?

test/fixtures/elastic-responses/example-get-response-object2.json
nelsonic: @SimonLab _many_ people I've worked with prefer a "_known state_" for their test/fixture data.
(_I personally prefer to use **real data** because it tells us quickly if something is "broken" ... see: https://github.com/numo-labs/lambda-ne-classic-package-provider/blob/3d2e649c87237d269cf62c8e2c12eb0e8a847d37/package.json#L7 where we create our fixtures from the **acutal** API/data before running our tests_)

nelsonic:
@SimonLab I _prefer_ dynamically creating fixtures, but its a _contentious_ area.
We might need to write a blog post discussing it and then re-visit. Please create an issue: https://github.com/dwyl/dwyl.github.io/issues 
Happy to merge for now. 👍

iteles:
@nelsonic Should we create this also as a `priority-5` open question in _this_ repo's backlog?

Issue #110, opened 2016-06-27
Joi validation search query with filter


SimonLab:
Add a validation for the filter parameters of the query string

Danwhy:
To do the validation slightly differently for the api vs the html, we're removing the joi validation from the config, and applying it to the requests individually

SimonLab:
resolved with #129

PR #129, opened 2016-07-01
Joi

Danwhy:
ref #110 
Removes validation from the config and does it manually on the parameters and queries as they are received in the request. This is because the jsonapi request and the html request accept different data types


nelsonic:
https://youtu.be/GVUTiuPuxMc?t=6s

SimonLab:
Good for me!

test/search-filters.test.js
```diff
@@ -739,3 +889,18 @@ testWithServer('Should accept multiple image_licences', (t, ctx) => {
     t.end();
   });
 });
+
+testWithServer('Should not accept array of multiple image_licences as json', (t, ctx) => {
+  t.plan(1);
```
nelsonic: do we need _both_ `t.plan(1);` _and_ `t.end();` ?

Danwhy: Possibly not, but [Alan suggested that we should](https://github.com/TheScienceMuseum/collectionsonline/pull/28#issuecomment-226163479)

nelsonic: Ok. cool.

Issue #111, opened 2016-06-27
Elasticsearch aggregations and filters search


SimonLab:
Use elasticsearch aggregations and filters to filter the results depending on the search

SimonLab:
The elasticsearch query has been implemented with #140

PR #140, opened 2016-07-04
build facets and filters

SimonLab:
# Ready for review

see  #111


nelsonic:
moar-tests

SimonLab:
@nelsonic do you mind to review this PR, thanks?

nelsonic:
@SimonLab I will take a second look tomorrow morning. (_please excuse delay..._)

lib/transforms/search-results-to-jsonapi.js
```diff
@@ -260,6 +260,137 @@ function createCountCategories (results) {
   return result;
 }
 
+/**
+* Create the facets object depending of the type of hte search (all, people, objects, documents)
+*/
+function createFacets (type, results) {
+  if (type === 'people') {
```
nelsonic: consider using a [`switch`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Statements/switch) ...?

also, given the format of the facet method you are invoking you could require an Object and invoke using a variable ... here's an example:

`facets.js`:
```js
module.exports = {
  'createDocumentFacets': (results) => {
    // your function here...
  },
  'createObjectFacets': (results) => {
    // your function here...
  } // ... etc.
}
```
Then when you `require` or `import` the `facets.js` you can invoke the method:
```js
const facets = require('./facets.js');
function createFacets (type, results) {
  return type ? facets['create' + type + 'Facets'](results) : {};
}
```

nelsonic:
Code looks good. left a couple of comments, but happy to merge. 👍

Issue #108, opened 2016-06-27
 boosting relevance


SimonLab:
some ideas:

- Plug Google Analytics with Elasticsearch - the more a document is clicked is more is relevant
- Log the search queries to know which ones are the most common and focus the relevance on these searches
- Document with picture are more relevant
- Document containing more data are more relevant

iteles:
2 days allocated to this in sprint 3, towards the end. Dependent on @jamieu providing details around what needs to be updated.

Danwhy:
In progress as #213

PR #132, opened 2016-07-04
Person visual-data fixes

Danwhy:
ref #106
* Uses un-comma-separated name for people
* Puts date of birth next to name
* Uses _industry_ and _based in_ for organisations, _occupation_ and _born in_ for people

Still to do: related objects/documents, but none exist in the database yet

PR #141, opened 2016-07-05
Search

Danwhy:
Simple update to the search function to make results marginally more relevant. Eg. searching 'charles babbage' gives Charles Babbage as the first result

ref #108


SimonLab:
Looks good, I'm merging it to resolve [the problem with Travis](https://github.com/TheScienceMuseum/collectionsonline/pull/132#issuecomment-230425781)

PR #134, opened 2016-07-04
Param object

SimonLab:
see #128

Issue #128, opened 2016-07-01
Create param and query object


SimonLab:
Create an object which group and abstract all the request.params and request.query. This object will help to build the elasticsearch facets:

```
{
 type: 'people',
 filters: {
   occupation: ['mathematician', 'poet'],
   place_born: 'London',
  ... 
 }
}
```

SimonLab:
done with #134

PR #142, opened 2016-07-05
uses queryparams object in client side ajax request #128

Danwhy:
Query params object created in #134 was not passed in to client side ajax request, so it was no longer working


SimonLab:
LGTM, @nelsonic?

nelsonic:
Yeah, looks good. merging. 👍

Issue #136, opened 2016-07-04
update url on new ajax search


SimonLab:
When a new search is executed with ajax we need to update the url with the right query parameters.

PR #145, opened 2016-07-05
Url update

### LINKED ISSUE #126: JS updating of results page needs to also update url
Otherwise when a user clicks though to a result page and then uses their browser to go back to the results, they are returned to the original (not latest) search query.

This may well be in hand, if so feel free to close this issue.

### LINKED ISSUE #143: Back button sometimes making json/ajax request
Sometimes pressing the back button to the search page returns the json search result, not the html. Need to look into why/how this is happening

### LINKED ISSUE #149: Toggle Facets without javascript
togglemenu

The Facets are display only by using javascript. So if javascript is not activated the facets won't be displayed.
Should we display automatically the facets when the javascript is disable?

Danwhy:
* Integrates home page search page into page.js for fast client side rendering: ref #48 

* Updates url when client makes a new request, and uses that url when client clicks through to new page and then uses back button: closes #126 and #136 

* This also seems to fix the problem we were having with json results being rendered instead of html on the back button press: fixes #143 

* The filters panel is now open by default, so users with javascript disabled can see it, then it is hidden with js as soon as the page is loaded: closes #149


SimonLab:
LGTM, @nelsonic?

nelsonic:
@SimonLab / @Danwhy yeah, code looks good.  🚀 
(_would prefer a few more tests, but happy to merge._)

Issue #162, opened 2016-07-08
Improving tests with real data in the fixtures


iteles:

SimonLab:
Is issue has been resolved with #169 . For each new feature we will add some copy of the real data to this fixtures.

PR #169, opened 2016-07-11
build fixture json data

SimonLab:
see  #162

Ready for review


SimonLab:
Close this PR because some tests have to be redefine. All tests are now updated on #171

PR #181, opened 2016-07-14
Related

Danwhy:
Gets related items from database

Returns related items in api response and html response. Currently set to 24 items, but will be variable with a query in the future.

@SimonLab Can you pull this down and check the tests run fine for you, then have a look over it please?


lib/sort-related-items.js
```diff
@@ -0,0 +1,23 @@
+const config = require('../config');
```
SimonLab: I think we could do the logic of the function by using aggregations on the request.

SimonLab:
We need to start adding more documentation to be able to follow the flow of the logic of the code. We also should add more jsdoc on the functions to help us to know quickly what is the role of these functions. Otherwise LGTM

lib/jsonapi-response.js
```diff
@@ -60,58 +59,90 @@ function getRelationships (data) {
...
+  return {
+    type: el.type,
+    id: el.id
+  };
+}
+
+function getIncluded (data, relationships, config, relatedItems) {
   if (relationships) {
-    const included = [];
+    var included = [];
```
nelsonic: there are a few changes from `const` to `var` but not _universally_ applied. is this _deliberate_ ?

Danwhy: Based on your previous suggestion not to use const for arrays/objects that are being added to, I've changed any I've noticed from before that were const

test/client/end-to-end.clienttest.js
```diff
@@ -9,6 +9,7 @@ module.exports = {
       .assert.urlEquals('http://localhost:8000/search?q=ada')
       .click('.resultcard')
       .assert.urlEquals('http://localhost:8000/people/smgc-people-38764')
+      .waitForElementVisible('.record-top__title', 1000)
```
nelsonic: @Danwhy you can set a _global_ value for Nightwatch to wait for e.g: https://github.com/dwyl/learn-nightwatch/blob/master/nightwatch.conf.BASIC.js#L23 that way you can leave the second argument out and nightwatch will wait the global amount as a maximum.

nelsonic:
LGTM! 🎉

Issue #189, opened 2016-07-15
Add documentation for page.js logic flow


SimonLab:
Our page.js ajax request logic can be refactor, see #188 #186 . At the moment it is also not quiet intuitive where and how to modify the code. For example I need to add a event listener on the input date to be able to filter on the dates but it took me a bit of time to understand the flow of the program and to be sure that I'm not breaking anything.

The documentation will resolve the above issue

SimonLab:
The wiki page is now created. The doc will be update at each new PR

PR #196, opened 2016-07-18
Pagejs refactoring

SimonLab:
see #189 
you can see the start of the documentation on [page.js](https://github.com/TheScienceMuseum/collectionsonline/wiki/Page.js) and [search](https://github.com/TheScienceMuseum/collectionsonline/wiki/Search)


client/lib/convert-url.js
```diff
@@ -1,23 +0,0 @@
-var QueryString = require('querystring');
```
SimonLab: This file was use to convert url between a json and url format for the ajax request. We only use a html to json format url which is simplified with the file to-json-url.js

SimonLab:
@nelsonic @Danwhy comments are welcome, thanks

client/middleware/error-404.js
```diff
@@ -1,12 +1,24 @@
+var Templates = require('../templates');
+
 module.exports = function (page) {
-  page(function (ctx, next) {
+  // page(function (ctx, next) {
+  //   next();
+  // });
+  page(render);
+  function render (ctx, next) {
     console.warn('Page not found', ctx.path);
```
nelsonic: Do we _always_ `console.warn` ?

client/routes/home.js
```diff
@@ -2,35 +2,30 @@ var svg4everybody = require('svg4everybody');
 var $ = require('jquery');
 var QueryString = require('querystring');
 var searchBox = require('../lib/search-box');
-var createQueryParams = require('../../lib/query-params.js');
-var getData = require('../lib/get-data.js');
+var data = require('../../fixtures/data');
```
nelsonic: is this _temporarily_ rendering _dummy data_?

client/middleware/error-404.js
SimonLab: I think we use console.war only to display which file couldn't be loaded and when a document of elasticsaerch doesn't belong to any of the main type (people, documents or object)

client/routes/home.js
SimonLab: yes currently there are no dynamic data on the home page

nelsonic: That's fine (_for now_) because it will load faster. 🐎

nelsonic:
Looks good. 👍

Danwhy:
LGTM

PR #208, opened 2016-07-22
Archive plus

Danwhy:
Now gets all archive children instead of just 10, and sorts them by identifier


lib/sort-related-items.js
```diff
@@ -26,6 +26,18 @@ module.exports = (related, type) => {
...
+      var aIdentifier = a.attributes.identifier;
+      var bIdentifier = b.attributes.identifier;
+
+      if (aIdentifier < bIdentifier) {
+        return -1;
+      }
+      if (aIdentifier > bIdentifier) {
+        return 1;
+      }
+
```
SimonLab: What is the type of identifier? Do you sort alphabetically? Are the identifier unique or do you need to add another condition where a.attributes.identifier === b.attributes.identifier?

SimonLab:
Just one question otherwise LGTM, @nelsonic?

lib/sort-related-items.js
Danwhy: The identifier is a string such as 'BAB/A', 'BAB/B', 'BAB/A/1', 'BAB/A/2', so they're being sorted alphanumerically. They should be unique, but there is a condition to return 0 if not, just in case.

nelsonic:
Looks good gents. Merging. :shipit:

Issue #222, opened 2016-07-28
Ajax request should be sent if the page is loaded for the first time


SimonLab:
The page for the person, document and object are called twice, one normal request and one request send with page.js. Check with the variable ctx.isInitialRender if we need to send the ajax request

jamieu:
Seems to be working much more smoothly now.

One (semi related issue) is #228 on Safari.

PR #224, opened 2016-07-29
initial render

SimonLab:
see #222 and the short explanation on the wiki [here](https://github.com/TheScienceMuseum/collectionsonline/wiki/Page.js#initial-render)


SimonLab:
@Danwhy ready for review, thanks

client/lib/get-data-item.js
```diff
@@ -0,0 +1,24 @@
...
+  })
+  .then(function (json) {
+    if (json.errors) return Promise.reject(json.errors[0]);
+    var data = JSONToHTML(json);
+    return cb(data);
+  })
+  .catch(function (err) {
+    console.error('Failed to find the item', err);
+  });
+};
```
Danwhy: This function is quite similar to https://github.com/TheScienceMuseum/collectionsonline/blob/master/client/lib/get-data.js
Maybe we should just use the one, and transform the data in the callback

Danwhy:
LGTM

PR #247, opened 2016-08-02
makes text/html default response if no media type given

Danwhy:
If no media type is given we're currently just sending back an empty reply.

The twitter bot that scrapes our site to get the data for the twitter cards has no media type, but expects html as a response, so this sends an html response if no media type is specified.

### DIFF
--- a/routes/archive.js
+++ b/routes/archive.js
@@ -4,7 +4,7 @@ const getArchiveAndChildren = require('../lib/get-archive-and-children');
 module.exports = (elastic, config) => ({
   method: 'GET',
   path: '/documents/{id}/{slug?}',
-  handler: (request, reply) => reply(),
+  handler: (request, reply) => HTMLResponse(request, reply, elastic, config),
   config: {
     validate: {
       query: archiveSchema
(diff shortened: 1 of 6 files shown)


routes/person.js
```diff
@@ -65,3 +38,31 @@ module.exports = (elastic, config) => ({
     }
   }
 });
+
+function HTMLResponse (request, reply, elastic, config) {
+  const data = {
+    page: 'person'
+  };
+
+  elastic.get({index: 'smg', type: 'agent', id: TypeMapping.toInternal(request.params.id)}, (err, result) => {
```
SimonLab: Maybe we could create a function for this elasticsearch request. It is almost the same code as in the json api handler.

routes/object.js
```diff
@@ -51,3 +30,24 @@ module.exports = (elastic, config) => ({
     }
   }
 });
+
+function HTMLResponse (request, reply, elastic, config) {
```
SimonLab: I prefer to create a new file for this function (and maybe we can reuse it on other endpoint). It allow us also to create a specific unit test for it. But it's just my preference and it's also fine for me the way it is.

routes/archive.js
```diff
@@ -4,7 +4,7 @@ const getArchiveAndChildren = require('../lib/get-archive-and-children');
 module.exports = (elastic, config) => ({
   method: 'GET',
   path: '/documents/{id}/{slug?}',
-  handler: (request, reply) => reply(),
+  handler: (request, reply) => HTMLResponse(request, reply, elastic, config),
```
SimonLab: So if hapi-negotiator doesn't manage to match the media type this will execute by default, that's right? I try to find if hapi-negotiator has a default handler if none of the media type is matched but couldn't find this option yet. Do you know if it exists? Otherwise it's ok!

Danwhy: No, I had a look and couldn't find an option either, that's why I did it like this. But yeah, by default it should now send html back, rather than an empty response

SimonLab:
@Danwhy just some questions and maybe some refactoring but mostly is based on my preferences and it's not mandatory so I let you decide what you prefer.

Issue #176, opened 2016-07-13
Add "approved web description" to object pages


jamieu:
This is stored in the index as **options.option1** (I know not very intuitive!)

When we have a "web description" (**options.option1**) it should overwrite the normal description on both HTML pages **and** the API.

{"query":{"match":{"identifier.value":"L2014-4158"}}}

Note: this is only for Object pages

SimonLab:
another way to get all the objects with the "option1" property defined:

```
curl /smg/object/_search -d
```
and the following payload

```
{
  "query":{
    "filtered": {
      "query": {
        "match_all":{}
      },
      "filter": {
         "exists" : { "field" : "options.option1" }
      }
    }
  }
}
```

SimonLab:
@jamieu the PR for this issue has been merged and the new version will be deployed soon. Can you please test the functionality and close this issue if everything is fine, thanks.

jamieu:
Looks good 👍

PR #253, opened 2016-08-04
grab approved description if defined for object

SimonLab:
see #176


jamieu:
Yep, that all looks good 👍

Issue #270, opened 2016-08-05
Use this image button is acting up / not working


jamieu:
The image button seems to have stopped working (not sure if this is @Danwhy working on Zooms?)

<img width="346" alt="screen shot 2016-08-05 at 13 30 27" src="https://cloud.githubusercontent.com/assets/91365/17436699/cda05926-5b10-11e6-93ca-0c1a695dbf49.png">

PR #276, opened 2016-08-08
makes cite button work 

Danwhy:
ref #270 

Button was using foundation's data toggles. Foundation's jquery function wasn't getting called on re-render. This fixes the button and removes dependency on foundation, for when we decide to get rid of jquery altogether


SimonLab:
LGTM

PR #296, opened 2016-08-12
initial relevance queries 

Danwhy:
Initial thoughts on improving the query to get better relevance:

Added `should:match_phrase` to the query for title, name and summary_value. This should provide an increased score for items that have the query terms close to each other in the title.

Boosting functions (`function_score:functions`): boosts the score of items that match certain terms, or have certain fields. eg. an item that has the `location.purpose`: `on display` will have its score boosted by 1.2, as will an item that has the field `options.options1` (the web description), while an item that has an image will be boosted by 1.4, and documents will be boosted by 0.8.

Boost numbers are subject to change, these are just initial values I was trying out to get some variance in results.


alanshaw:
@Danwhy could you add some information in your PR description to explain the changes you've made here and the affect they will have on the search results?

alanshaw:
I see "archive" result types are given a lower ranking although I think the [requirement from Jamie](https://github.com/TheScienceMuseum/collectionsonline/wiki/Ordering-and-relevance#archives) was for them to be lower if no particular type filter was applied - i.e. on the "all" tab. Could you implement this instead?

Danwhy:
The lower ranking should only apply to the 'all' tab anyway. If any other tab is selected, the archives either won't show up at all (objects, people), or the results will all be archives, in which case their scores will all stay the same relative to each other

alanshaw:
Looks like a great start

PR #298, opened 2016-08-15
adds link to search object type filter

### LINKED ISSUE #150: Facet values in body should link back to search results page
- MATERIALS
- TYPE
- CATEGORY

Danwhy:
Adds link from resource page to search results with 'object type' filter.

Last facet link from resource page, so closes #150

### DIFF
--- a/lib/transforms/json-to-html-data.js
+++ b/lib/transforms/json-to-html-data.js
@@ -142,7 +142,7 @@ function getDetails (resource, links) {
   const legal = getLegal(resource.data);
   const arrangement = getArrangement(resource.data);
   const notes = getNotes(resource.data);
-  const objectType = getObjectType(resource.data);
+  const objectType = getObjectType(resource.data, links);
   const taxonomy = getTaxonomy(resource);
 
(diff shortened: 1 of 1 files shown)


alanshaw:
👓

lib/transforms/json-to-html-data.js
```diff
@@ -328,15 +328,15 @@ function getNotes (data) {
...
+function getObjectType (data, links) {
   var value;
   if (data.type !== 'people' && data.attributes.name) {
     data.attributes.name.forEach(el => {
       if (el.primary) {
         value = el.value || null;
       }
     });
-    return value ? {key: 'type', value: value} : null;
+    return value ? {key: 'type', value: value, link: links.root + '/search/objects?filter[type]=' + value} : null;
```
alanshaw: `value` in URL needs `encodeURIComponent` in case it contains URI reserved characters like & for example.

Danwhy: 👍 I did that for the other links, will add it to this one

alanshaw:
LGTM

Issue #279, opened 2016-08-08
List view : add description text


jamieu:
Decision made: let's show up to 2 lines of description text (2/3rd's of the way across page, so that it look neat). Truncated on a word boundary with three eclipses ... if longer.

<img width="798" alt="screen shot 2016-08-08 at 14 25 05" src="https://cloud.githubusercontent.com/assets/91365/17481246/f65c3bc2-5d73-11e6-8029-064867f6bbc2.png">

SimonLab:
@jamieu do you also want to display the description on the mobile version or it will be too much information?

jamieu:
Good point, maybe add it in and see how it looks.

(sorry I know that's not a very strong decision!)

Jamie

alanshaw:
@SimonLab is complete now? Please close if true

jamieu:
Looks good, closing

PR #300, opened 2016-08-15
add description to list-view

SimonLab:
see #279

This PR add the [figcaption](https://github.com/TheScienceMuseum/collectionsonline/blob/0c946971fcb7ee3b3b77de4e46afb5dd2b6db991/lib/transforms/search-results-to-template-data.js#L75-L90) value to the list view display.

### DIFF
--- a/templates/partials/search/results-list.html
+++ b/templates/partials/search/results-list.html
@@ -16,6 +16,7 @@
         <p class="resultlist__description">{{this.category}}</p>
         <p class="resultlist__description">{{this.date}}</p>
         <p class="resultlist__description">{{this.occupation}}</p>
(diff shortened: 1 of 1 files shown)


alanshaw:
👓

alanshaw:
Not sure we can do ellipsis at the word boundary but we can certainly do ellipsis with `white-space: nowrap; overflow: hidden; text-overflow: ellipsis;` or is the plan to get Toby to look into this?

SimonLab:
On this version, I just reuse the description we display on the list-card view instead of creating a new  specific version with the ellipsis. Also the descriptions seem short enough and I'm not sure the ellipsis are necessary.

Issue #209, opened 2016-07-22
Simple means of creating static content pages (for footer links/FAQ pages)


jamieu:
Longer term the footer links can link to the main CMS where the pages can be edited.

In the short term:

- [ ] create simple / easily re-directable urls for key FAQ/help pages (@jamieu to supply).
- [ ] create a simple means of allowing SMG to edit these pages (a first pass at this could use a simple markdown format).
- [ ] create a simple means of pulling the content for these pages from an external source (ie. the main website) so that content can be edited without the need to redeploy the whole application. The though here is that SMG editorial staff could edit and upload a config file but wouldn't be capable of deploying a whole application from GitHub/Travis.

SimonLab:
@jamieu just to be sure, the idea is to be able to change easily the content of the pages linked in the footer:
![footersmg](https://cloud.githubusercontent.com/assets/6057298/17289989/1c5341b8-57d4-11e6-9ea6-4e5c00dca357.png)

We could create a separate object file for each of the pages like we did for the home page: [home page object](https://github.com/TheScienceMuseum/collectionsonline/blob/master/fixtures/data.js).

Then on your third point, the idea is to have an easy way for people from SMG to change these object files. One way of doing this is to have a login interface for the users to connect to a simple interface to update and save the file object. This requier to create an authentication system so maybe it will be more secure to build this system on an internal server. @Danwhy @nelsonic do you have any other ideas on how to do this?
@jamieu is my recap right?

jamieu:
Currently investigating which of these pages can live externally within the existing group website and simply be linked to. The bottom three links almost certainly fall into this category, along with the links to licensing, library and archive information.

Loath to go to to much effort here as mid-term we'll probably host most of these pages with the main CMS. They are unlikely to change often, but if the terms of a licence changes then it will be seen as important the the website reflects this.

What might be useful is a hybrid approach where we have default 'objects' for each footer page, much like the homepage. With a 'Link Title', 'Page Title' and 'Body text (with markup)' field for each. **But with the ability to specify a 'url' field which overrides the local copy**. If that's a fairly minor dev effort then lets go with that as it covers the basic use cases.

alanshaw:
Footer links can now be configured thanks to #301 - waiting on content from @jamieu for the pages that need it...

iteles:
@jamieu and Emily agreed to look at content for this in the coming sprint, from Monday 22nd August.

jamieu:
[...]

PR #301, opened 2016-08-15
build footers from json data

SimonLab:
see #209

### DIFF
--- a/client/routes/home.js
+++ b/client/routes/home.js
@@ -7,11 +7,17 @@ module.exports = function (page) {
   page('/', render, listeners);
 
   function render (ctx, next) {
-    var pageEl = document.getElementsByTagName('main')[0];
-    pageEl.innerHTML = Templates['home'](data);
-    // refresh the title of the page
-    document.getElementsByTagName('title')[0].textContent = data.titlePage;
-    next();
+    if (!ctx.isInitialRender) {
+      var pageEl = document.getElementsByTagName('main')[0];
+      data.footer = require('../../fixtures/footer');
+      data.footerBanner = require('../../fixtures/footer-banner');
+      pageEl.innerHTML = Templates['home'](data);
+      // refresh the title of the page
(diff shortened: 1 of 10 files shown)


alanshaw:
👓

alanshaw:
👓

client/routes/home.js
```diff
@@ -7,11 +7,17 @@ module.exports = function (page) {
...
+    if (!ctx.isInitialRender) {
+      var pageEl = document.getElementsByTagName('main')[0];
+      data.footer = require('../../fixtures/footer');
+      data.footerBanner = require('../../fixtures/footer-banner');
+      pageEl.innerHTML = Templates['home'](data);
+      // refresh the title of the page
+      document.getElementsByTagName('title')[0].textContent = data.titlePage;
+      next();
+    } else {
+      listeners();
```
alanshaw: This should be a call to next also, so we don't introduce a bug in the future when the `listeners` function also needs to use `ctx`.

alanshaw:
@SimonLab I'm not sure if the requirements have changed but as far as I can see this PR only allows the footer links to be easily configured to point at external URLs - in @jamieu's [last comment](https://github.com/TheScienceMuseum/collectionsonline/issues/209#issuecomment-236546216) he was looking for a combination of the two - a on-site page (and markdown based page text) for each footer item that can be overridden by an external URL if set.

Could you please let me know if this PR is meant to resolve the issue (because requirements have changed) or if it is simply a step towards completing it?

SimonLab:
At the moment we don't have any content or template for the pages linked on the footer. At the moment each link on the footer are described on the footer.js file and each link is defined with the url property. This url can target the future page on the website or an external page.
So I think I just need to have the list of external links otherwise a template to create the content of the pages, is this right or I missed some details of the feature?

alanshaw:
@SimonLab no thats all good - we're waiting on content...

Issue #278, opened 2016-08-08
Image zoom : Add open sea-dragon buttons for +/- and full screen


jamieu:
- [x] Need to pick icon ( @tobystokes )
- [x] Need to swap out standard OSD icons and place them under the new icons.
- [ ] True OSD full-screen needs icons too (need to find nice placement)
- [ ] Ensure they work on mobile viewports

Couple of points:

- true full screen icon needs to differentiate from "full to browser window" icon [should we enable the <> and >< icons with text?]
- extra icons should only display in zoom mode.
- mockup below is intentionally rough (need to pick good icons!)
- do we want to move the 'zoom' icons along so they separate from the open/close zoom icon?

<img width="845" alt="screen shot 2016-08-08 at 14 14 33" src="https://cloud.githubusercontent.com/assets/91365/17481129/38be3728-5d73-11e6-9a89-d09215647385.png">

tobystokes:
template was already done, see
https://github.com/TheScienceMuseum/collectionsonline/blob/master/templates/partials/records/record-imgpanel__controlbar.html#L24-L31

(but I expect that old `{{if osd}}` needs replacing)

I would say that EITHER the fullscreen OSD zoom button OR our standard full browser-height zoom button be available, but not both. Though there is not a lot of height difference in it, so no strong opinions on which one, and the same icon could be used for either

jamieu:
What's your thoughts on making it clear that there are effectively two
levels of zoom?

- zoom to browser window
- zoom to (true) full screen

tobystokes:
Don't have two levels, it's confusing.
Either stick with the standard zoom and hide the OSD one, or vice versa. but they do the same thing, practically.

jamieu:
Just to clarify I mean the really full-screen (outside the browser) full
screen.

But yes, maybe we should just jump straight to that and skip the
intermediate one. Maybe that's what you mean?

PR #302, opened 2016-08-15
adds buttons for openseadragon 

Danwhy:
ref #278

Adds new openseadragon buttons to bottom of the carousel. Gets rid of old zoom button, as it essentially did nothing except activate the viewer. New buttons all activate deep zoom viewer.

### DIFF
--- a/client/lib/openseadragon.js
+++ b/client/lib/openseadragon.js
@@ -1,21 +1,51 @@
 require('openseadragon');
 var $ = require('jquery');
 
-module.exports = function (imgUrl, ctx) {
-  if (!$('#openseadragon').length) return;
+module.exports = {
+  init: function (ctx, imgUrl, cb) {
+    imgUrl = imgUrl || $('.slick-active img')[0].src;
+    if (!$('#openseadragon').length) return;
 
-  ctx.viewer = OpenSeadragon({
-    id: 'openseadragon',
-    // element: $('.record-imgpanel__singleimg'),
-    prefixUrl: '/assets/img/openseadragon/',
-    showZoomControl: true,
-    // just using an example static image, will be swapped for some magic tiles
-    tileSources: {
-      type: 'image',
-      url: imgUrl
+    if (!ctx.viewer) {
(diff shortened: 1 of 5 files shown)


alanshaw:
👓

client/lib/openseadragon.js
```diff
@@ -1,21 +1,52 @@
...
-  ctx.viewer = OpenSeadragon({
-    id: 'openseadragon',
-    // element: $('.record-imgpanel__singleimg'),
-    prefixUrl: '/assets/img/openseadragon/',
-    showZoomControl: true,
-    // just using an example static image, will be swapped for some magic tiles
-    tileSources: {
-      type: 'image',
-      url: imgUrl
+    if (!ctx.viewer) {
```
alanshaw: When would `ctx.viewer` not be `null`?

client/lib/openseadragon.js
```diff
@@ -1,21 +1,52 @@
...
+        if (e.fullScreen) {
+          e.eventSource.gestureSettingsMouse.scrollToZoom = true;
+        } else {
+          e.eventSource.gestureSettingsMouse.scrollToZoom = false;
+        }
+      });
+
+      $('#openseadragon .pic').hide();
+
+      ctx.save();
```
alanshaw: Ah, I see - ignore that last comment!

client/lib/openseadragon.js
```diff
@@ -1,21 +1,52 @@
...
+        },
+        zoomInButton: 'osd-zoomin',
+        zoomOutButton: 'osd-zoomout',
+        fullPageButton: 'osd-fullpage',
+        homeButton: 'osd-home',
+        toolbar: 'openseadragon-toolbar',
+        gestureSettingsMouse: {
+          scrollToZoom: false
+        }
+      });
```
alanshaw: Does this actually work? - if we save the context and then navigate away and back to the page the DOM element(s) will be different, does `OpenSeadragon` deal with this....?

SimonLab:
Done with the new carousel

PR #303, opened 2016-08-16
Format example responses so humans can view them

alanshaw:
* Makes the examples human readable
* Makes it easier to see what has changed when the index is updated


SimonLab:
LGTM!

PR #317, opened 2016-08-17
Clean up and document npm test scripts

alanshaw:


SimonLab:
:+1: nice! I think we over complicated the tests when we added the --data parameter on the command line.

Issue #217, opened 2016-07-26
Relevance - Synonyms


Danwhy:
It looks like the ideal way of dealing with synonyms is to use the [synonym token filter](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-synonym-tokenfilter.html), which has to be done at index-time.

We could do this in another way for now, using some kind of json file and comparison at query time, but it would obviously not be as efficient. @jamieu thoughts?

SimonLab:
The idea for the synonyms is to update the list of match keywords without reindexing all the database.
I manage to test the synonyms search:
- define a synonyme filter and analyzer
- apply the the synonyme analyzer to the fields by defining the mapping (it's the only time when we need to reindex the database to be able to apply the synonyms)
- search with synonyms

From now we can update the list of synonyms and we only need to update the settings of the index (so no need to reindex) to be able to search at query time. I don't think we need to apply the synonyms at index time:
![query-time-analyzer](https://cloud.githubusercontent.com/assets/6057298/17706220/155af742-63d3-11e6-8a6e-c9059b74e86c.png)

However this kind of search synonyms only work well for only one word synonym. I'm looking now how to search for multiple word synonym (https://www.elastic.co/guide/en/elasticsearch/guide/current/multi-word-synonyms.html).

@jamieu do you have an exemple of synonym that we can try to imlement? Do you also think that we will need to work with multiword synonym?

jamieu:
Two off the top of my head...

"trains, locomotives, rolling stock" -> locomotives
"orthodontics, dentistry, teeth, dental" -> dentistry

SimonLab:
The script to update the list of synonyms is working on my localhsot elasticsearch, however I can't test it on the real database as we don't have the rights to modify the settings of the database:
![forbiden](https://cloud.githubusercontent.com/assets/6057298/17775121/bdc6cb14-654e-11e6-848e-b7d73f6fb3d0.png)

@jamieu would it be possible to have a specific user with the rights to change the settings?

PR #321, opened 2016-08-18
Synonyms

SimonLab:
# not ready yet

see #217


nelsonic:
@SimonLab is this still "_in scope_"?

SimonLab:
@nelsonic still in scope but I think I would prefer to create a new repo for this specific feature. Updating a list of synonyms on the settings of Elasticsearch is more a maintenance task and I don't want to add this type of feature on top of the application, does this sounds good?

PR #326, opened 2016-08-19
User Improved Search Results

Danwhy:
### To be merged after #346

Adds basic logging of how many times users click on search results.

Stores number of clicks in leveldb and provides prefixes to keys to easily identify 'count groups', so results with counts between certain values (say 10000 and 100000) are grouped together for easy extraction from leveldb, and then given a score boost.


lib/scores.js
```diff
@@ -0,0 +1,114 @@
+const Async = require('async');
```
SimonLab: A wiki page for this file will be great

SimonLab:
Just some doc to add to explain how we define the boost depending on the count, otherwise LGTM

Danwhy:
Added some documentation here https://github.com/TheScienceMuseum/collectionsonline/wiki/User-Improved-Search-Results

lib/scores.js
```diff
@@ -0,0 +1,114 @@
...
+const level = require('../lib/level.js');
```
nelsonic: @Danwhy bearing in mind that this ties the _score_ to a single server I would _recommend_ you consider asking @jamieu if we can use Redis (ElastiCache): http://docs.aws.amazon.com/AmazonElastiCache/latest/UserGuide/Clusters.Create.Redis.CON.html to ensure _horizontal scalability_ ...

Danwhy: @nelsonic  We definitely were thinking about using Redis eventually, Alan just suggested LevelDB as a simple way to get started without having to worry about deploying anything new. We've got an open issue reminding us to consider other options when scaling: https://github.com/TheScienceMuseum/collectionsonline/issues/339, I'll add the Redis suggestion

nelsonic:
@Danwhy the code looks (_really_) good. 
_However_ I would _recommend_ that you use Redis instead of LevelDB to ensure that on-instance data is not lost in a crash and we can add nodes to the ELB cluster...
I _know_ that _initially_ we won't "_need_" to have multiple Node.js nodes ... but it would _good_ to have "_failover_" ...
Thoughts? Was this discussed on an issue? (_sorry to give you more work..._)

lib/scores.js
nelsonic: Cool. ok. as long as the team are _aware_ that we either need to implement "_sync_" between the node.js servers (_if they use LevelDB_) or use ElastiCache to store the data, I'm happy with either. 👍

Danwhy:
I'd be very happy to use Redis for this. Will just have to ask Jamie about getting it up and running

nelsonic:
@Danwhy let me know if you need anything from me to make progress with this. 👍

Danwhy:
@nelsonic So we've decided to leave this for now, as @jamieu is going to look into using the google analytics for the tracking, and then we can use that to boost the relevance. The code may be useful in the future though, so I'd like to keep the branch.  
I'll add a link to the branch in the relevant issue and close this pr.

jamieu:
I agree, leave for now and close issue, but keep the branch.

I can already see how we might use / build upon this in future for
scenarios where google analytics doesn't cut it.

Issue #335, opened 2016-09-01
"clear all filters" link is broken


SimonLab:
the query parameter q looks empty which produce a wrong url

SimonLab:
resolved with #336

PR #336, opened 2016-09-01
delete unsused clear filter function

SimonLab:
see #335


SimonLab:
@Danwhy do you mind to have a look at this small PR, thanks

Issue #338, opened 2016-09-02
Tests should be explicit about the fixture they need


iteles:
The test should be in charge of what it's getting, not using a generic database object

More important to just be able to remove tests - repetition in tests is ok and doesn't matter.

SimonLab:
done with #346

PR #346, opened 2016-09-05
Test Refactoring

Danwhy:
ref #338 

Simplifies tests by cutting out poorly managed fixtures, and using real database for now.

Using Sinon to stub elasticsearch/some other functions to allow for easy testing of errors. We could also use this in the future to return simpler fixtures, which would allow for testing offline, but as we have to be online for the nightwatch tests, that's not a huge priority.

Now with 100% coverage.

### DIFF
--- a/test/search-filters.test.js
+++ b/test/search-filters.test.js
@@ -907,39 +907,6 @@ testWithServer(file + 'Should not accept array of multiple image_licences as jso
   });
 });
 
-// testWithServer(file + 'Should return an error if the elasticsearch search function fail', {}, (t, ctx) => {
-//   t.plan(1);
-//
-//   const htmlRequest = {
-//     method: 'GET',
-//     // mock-database trigger an error if the query parameter is equal to "error"
-//     url: '/search?' + QueryString.stringify({ q: 'error' }),
-//     headers: { Accept: 'text/html' }
-//   };
-//
-//   ctx.server.inject(htmlRequest, (res) => {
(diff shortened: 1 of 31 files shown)


test/search-filters.test.js
```diff
@@ -907,41 +907,41 @@ testWithServer(file + 'Should not accept array of multiple image_licences as jso
...
-  };
-
-  ctx.server.inject(htmlRequest, (res) => {
-    const response = JSON.parse(res.payload);
-    t.equal(res.statusCode, 400, 'Status code was as expected to 400');
-    t.equal(response.errors[0].title, 'Bad Request', 'the title of the error is Bad Request');
-    t.end();
-  });
-});
+// testWithServer(file + 'Should return an error if the elasticsearch search function fail', {}, (t, ctx) => {
```
SimonLab: Can we delete the commented tests below?

Danwhy: Yep, sorry must have missed those

SimonLab:
I think we can also delete the [searchName property](https://github.com/TheScienceMuseum/collectionsonline/blob/168c25b6dcf3f6f0c1c35728f1a8affc0400b879/lib/search.js#L155) as we don't use it anymore.  The searchName property is also defined for the other type of searches so we can delete it for them too.

SimonLab:
LGTM!, @nelsonic ?

nelsonic:
@Danwhy really _stoked_ you made the time for this! 🎉

Issue #272, opened 2016-08-05
Archive people / names : we don't want to do the comma swap on Archive / AdLib names


jamieu:
(apologies not really a bug, more a fix/tweak - but didn't have a 'fix' or 'tweak' label!)

The behaviour/logic is currently correct for Mimsy people, but for archives we want to keep people and company names as they appear in the index.

@jamieu may need to revisit who we store / extract the names from AdLib as currently some of them are a bit odd / overloaded 

See: John Pease - Quaker Industrialist
- http://archives.sciencemuseumgroup.ac.uk/Details/archiverail/110002901
- /people/smga-people-260

<img width="1089" alt="screen shot 2016-08-05 at 13 38 07" src="https://cloud.githubusercontent.com/assets/91365/17436842/e660f01e-5b11-11e6-86a6-0d84a1d638b6.png">

<img width="736" alt="screen shot 2016-08-05 at 13 39 35" src="https://cloud.githubusercontent.com/assets/91365/17436880/23084422-5b12-11e6-91b3-f0970bf7cf96.png">

PR #355, opened 2016-09-07
Undo Normalisation of Names from AdLib System

Danwhy:
ref #272


SimonLab:
LGTM

nelsonic:
looks good. 👍

iteles:
@Danwhy Did we capture these changes in the wiki with a link to the line number so they can be easily backed out if/when the data is fixed and the code becomes unnecessary?

Issue #185, opened 2016-07-14
JSON-LD (in HTML markup)


jamieu:
Object (or a subset where identifiable such as a Photograph), Documents, People, Organisations, Locations and Places pages should expose their data using JSON-LD and the relevant schema.org classes.

https://developers.google.com/schemas/formats/json-ld
https://schema.org/Museum https://schema.org/object https://schema.org/Photograph https://schema.org/location https://schema.org/Place https://schema.org/Person https://schema.org/Organization

PR #366, opened 2016-09-08
adds JSONLD to HTML 

Danwhy:
ref #185 
Creates a json object based on keys from [schema.org](https://schema.org/docs/full.html),
then sends that with the rest of the data to be rendered in the head of the page


SimonLab:
LGTM, @nelsonic?

nelsonic:
Yeah, looks great! 👍

Issue #329, opened 2016-08-19
SEO-friendly URLs


iteles:
Improving the slug by adding a `/name-of-thing-on-this-page` to the URLs.

iteles:
Milestone cleared as time moved out of this sprint and into future sprints.

SimonLab:
Need to change the urls of the [sitemap](https://github.com/TheScienceMuseum/collectionsonline-sitemap/blob/bd38a646682e8b48717dde7584838cfbad083cb3/lib/hit-to-sitemap-entry.js#L8) to add the slug

jamieu:
Nice 👍

PR #367, opened 2016-09-08
add slug to the url

SimonLab:
see #329


jamieu:
Does the base url remain the canonical version? be good if any slug was just a 302 redirect to the base url so that i) we don’t get penalised by Google for serving the same content via multiple urls ii) if the name changes we don’t lose the canonical / permanent url.

Good point re. what to put in the sitemap, ideally we'd have Google index pages with the slug, but treat the base url as canonical. Although not sure it's possible to "have our cake and eat it too". I think we have to pick one or the other. Any thoughts?

SimonLab:
The slug is at the moment optional so the base url (without the slug) is still displaying the right data:

http://localhost:8000/people/smgc-people-36993/babbage-charles
http://localhost:8000/people/smgc-people-36993

are displaying the info page of Babbage.

 We can redirect to the base url but this will display the url without the slug in the address bar. Does this contradict why we build this feature as the users won't be able to guess what a page is based on the slug?

We might be able to add both url in the sitemap but I'm not sure if it's the right way to do it or if it's going to "confuse" Google? I will create a new issue for this

routes/person.js
```diff
@@ -8,7 +8,7 @@ const sortRelated = require('../lib/sort-related-items');
 module.exports = (elastic, config) => ({
   method: 'GET',
   path: '/people/{id}/{slug?}',
-  handler: (request, reply) => HTMLResponse(request, reply, elastic, config),
+  handler: (request, reply) => reply.continue(),
```
Danwhy: Do we want to change this yet, or wait until we're checking the user agent?

SimonLab: Yes let's wait, I need more thoughts on how to manage the content negociation with the user agent

SimonLab:
old PR replaced by #492

Issue #382, opened 2016-09-14
Only show category (name) on object pages, not the museum code.


jamieu:
This should just be "Computing and Data Processing" (category.name)

[ideally we'd do this in the model/controller so that they appear this way in the API as well] 

<img width="526" alt="screen shot 2016-09-14 at 12 15 26" src="https://cloud.githubusercontent.com/assets/91365/18510260/f395c00c-7a74-11e6-8e59-7c6625d1f3c1.png">

SimonLab:
The api version of an object page display more details for the category compare to the html version:
```
      "categories": [
        {
          "museum": "SCM",
          "name": "Art",
          "value": "SCM - Art"
        }
```
@jamieu do we want to limit this result for the api too?

jamieu:
Actually leave it like that, it makes sense in the API

:-)

jamieu:
@SimonLab Can we also change this on the SERP results

<img width="300" alt="screen shot 2016-09-16 at 13 01 11" src="https://cloud.githubusercontent.com/assets/91365/18585244/b68242d6-7c0d-11e6-8f13-9d3a7e9d99bc.png">

PR #386, opened 2016-09-15
only display name of the category in object page

SimonLab:
see  #382


nelsonic:
looks good. thanks for updating that. 👍

PR #389, opened 2016-09-16
lazy load images

Danwhy:
Lazy loads images and transitions them from transparent to opaque, to avoid images later in the carousel from appearing before the first image, and briefly causing flashing/overlapping images


nelsonic:
yeah, LGTM2. 👍

iteles:
Slight coverage drop seems to have come from this PR

Issue #362, opened 2016-09-08
Split object names and materials on comma (,) at point of display 


jamieu:
Related to https://github.com/TheScienceMuseum/collectionsonline/issues/361

Split object names and materials on comma (,) at point of display and link back to search separately so each term is it's own link (back to the SERP's page via the relevant filter field).

Ideally these would be stored in the index or source systems as an array of values, but in the intern can we split them as they are retrieved from the index and **link them separately** (as the combined string is pretty useless in terms of letting users continue on their search journey).

I realise this may result in false-positives (ie. a expected/correct comma in an material name or object name), if anyone is aware of where this might cause serious issues then do flag it up.

Issue #361, opened 2016-09-08
Split occupations on comma (,) at point of display


jamieu:
Related to https://github.com/TheScienceMuseum/collectionsonline/issues/362

Ideally these would be stored in the index or source systems as an array of values, but in the intern can we split them as they are retrieved from the index and **link them separately** back to the SERP's page via the occupation filter (as the combined string is pretty useless in terms of letting users continue on their search journey).

I realise this may result in false-positives (ie. a expected/correct comma in an occupation), if anyone is aware of where this might cause serious issues then do flag it up.

<img width="1040" alt="screen shot 2016-09-08 at 13 23 20" src="https://cloud.githubusercontent.com/assets/91365/18349126/713d36dc-75c7-11e6-895f-5add050d0c23.png">

PR #396, opened 2016-09-21
Serp links

Danwhy:
ref #362 #361
Adds links back to the search results page using individual terms rather than concatenated string of occupations/materials/types


nelsonic:
Looks great. thanks @Danwhy  👍 
❤️ ✅ 🎉

Issue #401, opened 2016-09-28
Error when selecting a 'materials link' on the object page 


jamieu:
page: /search/objects?filter[material]=cast%20iron,
link: /search/objects?filter[material]=cast%20iron,

Error:

<img width="725" alt="screen shot 2016-09-28 at 12 40 00" src="https://cloud.githubusercontent.com/assets/91365/18912124/b35781e2-8578-11e6-93b5-4da3a7868b8b.png">

jamieu:
Related to https://github.com/TheScienceMuseum/collectionsonline/issues/409

jamieu:
Seems to have fixed the comma issue, some values still not split in index (but have flagged that in https://github.com/TheScienceMuseum/collectionsonline/issues/409)

PR #404, opened 2016-09-28
adds filters for materials #401

Danwhy:
ref #401 
Adds materials to the aggregation/filter queries. 
Uses an array for the query parameters, so we can now select multiple materials if the object has more than one


nelsonic:
Yeah, code looks good. 👍 
(_nice one @Danwhy!_)

Issue #160, opened 2016-07-08
SMG to supply allowed list of relationships and association types


iteles:
This is the list of meaningful associations that will need to be kept in a JSON file and displayed if they are encountered (e.g. son, maker, etc)

iteles:
@jamieu to provide the top 20 temporary ones for now in this sprint (list will need to be modifiable with something like a JSON config file)

jamieu:
We should ideally store these values in a JSON config file so that they can be edited/updated over time/

## Makers

Default to "**Maker:**" (under title) _unless_ relationship is one of the following, 
	
- artist	
- publisher	
- printer	
- photographer	
- designer	
- manufacturer	
- author	
- engraver	
- supplier	
- inventor	
- inventor type	
- patentee	
- modifier	
- illustrator	
- preparer	
- specifier	
- lithographer	
- commissioner	
- issuer	
- creator	
- medallist	
- attributed	
- signatory	
- distributor	
- developer	
- railway company	

## Users

These should appear in the "Related people" block on the right hand side (rather than at the top of the page with makers). They should appear _without_ a relationship label _unless_ the relationship is one of the following, in which case the relationships should appear in brackets after the name.
	
- supplier	
- user
- preparer
- collector
- patentee
- collector
- inventor
- depicted
- sitter
- subject
- owner
- recipient

NOTE: See /objects/smgc-objects-205732 for a good example record:
NOTE: First letter of Maker or User/Association capitalised in all cases.

PR #405, opened 2016-09-29
approved values of maker and user

Danwhy:
ref #160, #418 
Suppresses role of maker/user/relation if not an approved value


lib/transforms/get-values.js
```diff
@@ -140,6 +141,9 @@ module.exports = {
           key = getNestedProperty(el, 'attributes.role.value') || getNestedProperty(el, 'attributes.role.type');
           value = getNestedProperty(el, 'attributes.summary_title') || null;
           link = getNestedProperty(el, 'links.self');
+          if (makersList.indexOf(key) === -1) {
+            key = 'maker';
```
SimonLab: Just to clarify, if the type of the maker is not in the list we are defining the type "maker" by default, is that right?

Danwhy: Yeah that's right

nelsonic:
pesky merge conflicts! (_please re-assign when resolved, thanks!_)

nelsonic:
@Danwhy merge conflicts ... 😞  (_please resolve and re-assign_)
(_code looks good tho..._)

nelsonic:
@Danwhy thanks for fixing. 👍 
(_@SimonLab thanks for doing initial review ... code great as always!_)

PR #407, opened 2016-09-30
coverage

Danwhy:
Adds some tests and deletes some dead code to increase coverage back to 100%

### DIFF
--- a/lib/facets/aggs-documents.js
+++ b/lib/facets/aggs-documents.js
@@ -15,18 +15,6 @@ module.exports = function (queryParams) {
             terms: { field: 'fonds.summary_title' }
           }
         }
-      },
-      organisation: {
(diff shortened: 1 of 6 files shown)


lib/facets/aggs-documents.js
```diff
@@ -15,18 +15,6 @@ module.exports = function (queryParams) {
             terms: { field: 'fonds.summary_title' }
           }
         }
-      },
```
Danwhy: There aren't any documents in the index that have an organisation, and I'm not sure if there ever were or will be

SimonLab: ok sounds good, if the index changes we will add the organisation facets if it's still necessary, but for now let's delete the unused code.

Issue #402, opened 2016-09-28
Incorrect counts on SERPs page (for type filter)


jamieu:
Following the 'steam locomotives' link ( /search/objects?filter[type]=steam%20locomotive ) from Rocket (/objects/smgc-objects-26704) leads to a SERPS page with incorrect counts.

ie. 83 'Objects' but 23668 'All' 

<img width="747" alt="screen shot 2016-09-28 at 12 45 02" src="https://cloud.githubusercontent.com/assets/91365/18912272/740b2c0e-8579-11e6-9439-24f81ba8315b.png">

jamieu:
Looks good 👍

Issue #265, opened 2016-08-05
Odd counts being returned


jamieu:
Some of the counts are coming back with unusual responses (e.g. Search “Rocket” and filter by poster, count increases from 16 to 17); (e.g. Search “Computer” and filter by analogue, count increases from 10 to 11).

Is this an issue with the logic or ES sharding?

SimonLab:
I think it's an issue with ES sharding, I'm investigating if there are any logs or errors:
see #137

jamieu:
Not seeing this anymore, so closing issue.

PR #411, opened 2016-10-02
Filter refactor

Danwhy:
ref #402, #265 
### To be merged after #404 

Refactors the filters, facets and aggregations so people/objects/documents are all using the same filters. This fixes some issues where counts were not correct or changing because certain filters we had applied were ignored if we switched from say 'objects', to 'people'.


nelsonic:
👀

nelsonic:
Very clean. Love the simplification. Thanks @Danwhy 👍 
❤️  ✅  🎉

PR #428, opened 2016-10-05
No results view

### LINKED ISSUE #421: Suppress the filter on the 'no results' page
Hopefully a small tweak, but on the 'no results' page could we:

- suppress the 'view icons' 
- suppress the 'refine search' block _(note: I can't think of a use-case where a filter would be selected and still display no results, but if you can, then maybe add in a check to "only suppress if nothing is checked")._

screen shot 2016-10-03 at 19 19 40

SimonLab:
resolved #421


Danwhy:
Looks good. Would be good to hide the text/button that says 'you have filters applied', 'clear filters' using this logic though, if there aren't any filters selected

Issue #399, opened 2016-09-28
Submitting an empty query


jamieu:
Could we just show the whole collection in the search results (for users to filter down) if not query is entered. Looks like we can already do this in some paces?

Rather than just showing the error below?

<img width="301" alt="screen shot 2016-09-28 at 12 32 08" src="https://cloud.githubusercontent.com/assets/91365/18911987/f01e019c-8577-11e6-80af-c036c343ff2a.png">

SimonLab:
This should have the same as doing a no search query

SimonLab:
resolved with #440

PR #440, opened 2016-10-06
allow empty search which return all the results

SimonLab:
see #399

[This code](https://github.com/TheScienceMuseum/collectionsonline/blob/f2fb6e8a3c05033e509d95424631da03894c306d/lib/search.js#L157-L159) will catch the empty query and return all the results.


Danwhy:
:+1:

PR #462, opened 2016-10-07
Disabled tab 430

tobystokes:
n.b. to reviewer, that js works, but if there is a nicer way to do it, let me know!

### DIFF
--- a/templates/partials/global/search-nav.html
+++ b/templates/partials/global/search-nav.html
@@ -1,19 +1,27 @@
-<a class="searchtab{{#if isFilterAll}} searchtab--active{{/if}}" href="{{links.filterAll}}">
+<a class="searchtab{{#if isFilterAll}} searchtab--active{{/if}}"
+  {{#unless filterAllDisabled}}href="{{links.filterAll}}"{{/unless}}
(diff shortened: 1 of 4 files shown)


templates/partials/global/search-nav.html
```diff
@@ -1,19 +1,27 @@
-<a class="searchtab{{#if isFilterAll}} searchtab--active{{/if}}" href="{{links.filterAll}}">
+<a class="searchtab{{#if isFilterAll}} searchtab--active{{/if}}"
+  {{#unless filterAllDisabled}}href="{{links.filterAll}}"{{/unless}}
+  {{#if filterAllDisabled}}aria-disabled="true"{{/if}}>
```
SimonLab: the two conditions above can be done with a if else:
```
{{#if filterAllDisabled}}
  aria-disabled="true"
{{else}}
  href="{{links.filterAll}}"
{{#if}}
>
```
@tobystokes do you have any preference?

tobystokes: no prefs, I just didn't think handlebars did {{else}}.

Issue #420, opened 2016-10-03
Odd behaviour with object filters


jamieu:
Apologies if this is already fixed, but seeing some odd behaviour with the object types filter. 

If you select 'poster' and then 'aeroplanes' I'd expect all "posters of aeroplanes" (ie. objects with both those values) and those two checkboxes to remain checked. The only other values shown at that point would be a third (or more) tag assigned to objects in the result set, whose first and second tag was "posters" and "aeroplanes".

<img width="729" alt="screen shot 2016-10-03 at 18 08 53" src="https://cloud.githubusercontent.com/assets/91365/19046292/83d8f61a-8994-11e6-9310-fc0c5de7d84f.png">

<img width="646" alt="screen shot 2016-10-03 at 18 09 07" src="https://cloud.githubusercontent.com/assets/91365/19046287/7fd587f4-8994-11e6-8c54-a45ab089674b.png">

Danwhy:
So this works with the 'material' filters, we just need to apply that same logic to the others. Should be quite a simple fix

jamieu:
Here is another example, 

1) User selects:
[x] therapeutics [injection] 221

2) You'd now expect to see everything filtered by therapeutics
[x] therapeutics [injection] _221_
[] hypodermic syringes _186_
[] bottle _1_
[] syring _1_

2) User now selects 'hypodermic syringes' you'd expect to see both  'hypodermic syringes' and "therapeutics " checked with the remaining values reflecting objects that have a 3rd tag.
[x] therapeutics [injection] _221_
[x] hypodermic syringes _186_
[] bottle _1_
[] syring _1_

In short the values should reflect the result set to the left.

See #244 

<img width="386" alt="screen shot 2016-10-05 at 13 04 34" src="https://cloud.githubusercontent.com/assets/91365/19112365/5ed95bca-8afc-11e6-8d51-1363765e8b29.png">
<img width="407" alt="screen shot 2016-10-05 at 13 04 28" src="https://cloud.githubusercontent.com/assets/91365/19112362/5eceaea0-8afc-11e6-83e4-51af53c7e693.png">
<img width="424" alt="screen shot 2016-10-05 at 13 04 21" src="https://cloud.githubusercontent.com/assets/91365/19112364/5ed1e66a-8afc-11e6-9aaf-1a66f5b8b776.png">

jamieu:
Fixed 👍

PR #412, opened 2016-10-03
Comma

Danwhy:

PR #469, opened 2016-10-10
Multiple filters

Danwhy:
### To be merged after #412

ref #420 
Allows multiple filters of the same facet to be selected when objects have more than one, by using an array for the filter instead of a single value.

I've also factored out the filter creation function from `aggs-all`, as it was almost identical to the filter creation function in `create-filters`


nelsonic:
_Nice_ work @Danwhy  👍

Issue #409, opened 2016-09-30
Comma being included in materials href


jamieu:
Related to https://github.com/TheScienceMuseum/collectionsonline/issues/401 and https://github.com/TheScienceMuseum/collectionsonline/issues/362

<img width="532" alt="screen shot 2016-09-30 at 16 18 46" src="https://cloud.githubusercontent.com/assets/91365/18996796/9a29e266-8729-11e6-9922-a73893041116.png">

<img width="374" alt="screen shot 2016-09-30 at 16 15 32" src="https://cloud.githubusercontent.com/assets/91365/18996689/3320120c-8729-11e6-85b8-39c29af3a93b.png">

jamieu:
Related to https://github.com/TheScienceMuseum/collectionsonline/issues/401

jamieu:
Still seeing this issue, but assume it's because the values haven't been split correctly in the index, so assign an "index issue" tag. Front end seems fine.

`"materials": [ "whole, metal" ],`

<img width="738" alt="screen shot 2016-10-31 at 16 11 49" src="https://cloud.githubusercontent.com/assets/91365/19861525/c1e3ae2c-9f84-11e6-81c2-2066652c9f11.png">

PR #471, opened 2016-10-11
separates comma from value in details 

Danwhy:
ref #409 
Moves comma logic to handlebars rather than server to avoid it showing in the url

### DIFF
--- a/lib/transforms/json-to-html-data.js
+++ b/lib/transforms/json-to-html-data.js
@@ -219,10 +219,7 @@ function getAccession (data) {
 function getMaterials (data) {
   var values;
   if (data.attributes.materials) {
-    values = data.attributes.materials.map((material, i, arr) => {
-      if (i < arr.length - 1) {
-        material += ',';
-      }
+    values = data.attributes.materials.map(material => {
       return {
         value: material,
         link: `/search/objects?filter[material]=${material}`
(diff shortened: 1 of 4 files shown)


lib/transforms/json-to-html-data.js
```diff
@@ -220,9 +220,6 @@ function getMaterials (data) {
   var values;
   if (data.attributes.materials) {
     values = data.attributes.materials.map((material, i, arr) => {
```
SimonLab: do we still need the i and arr parameters?

templates/helpers/comma.js
```diff
@@ -0,0 +1,6 @@
+module.exports = (item, parent) => {
+  if (parent.value[parent.value.length - 1].value !== item.value) {
+    return ',';
+  }
+  console.log(item, parent);
```
Danwhy: definitely not, sorry

lib/transforms/json-to-html-data.js
SimonLab: I think there is the same thing a bit under in the code with ```data.attributes.component.forEach((el, i, arr) => {``` I don't think we use i and arr so we can clean them too?

SimonLab:
@nelsonic I just add my comments for this PR otherwise looks good, do you want to have a look?

Danwhy:
@simonlab @nelsonic I have updated this. For some reason github doesn't tell you a diff is outdated anymore

nelsonic:
@SimonLab thanks for doing that initial review. 😉 
@Danwhy thanks for responding & updating.  ✅ 
no further comments. (_looks good_) :shipit:

Issue #449, opened 2016-10-06
Place filters not working correctly


jamieu:
This might be fixed in one of the current branches, but if not..
- If you go to the “Rocket” /objects/smgc-objects-26704
- Select “Newcastle upon Tyne” as a search filter
- This returns no results (even though it is selectable) and you have just come from a result

<img width="815" alt="screen shot 2016-10-06 at 13 29 41" src="https://cloud.githubusercontent.com/assets/91365/19152224/f70672a8-8bc8-11e6-99c3-52c8b8059460.png">

jamieu:
Resolved 👍

Issue #448, opened 2016-10-06
Default filters displayed after selecting a value


jamieu:
This might be fixed in one of the current branches, but if not..
1.       Go to a person record
2.       Select the place
3.       You return to a filter page which includes all the default filters (as if you had come to the /search page without a query)

ie. Joseph Corbett: /people/smgc-people-45804

![unknown-4](https://cloud.githubusercontent.com/assets/91365/19152162/a4ed7c32-8bc8-11e6-8278-2e0374aad67a.png)

jamieu:
👍

Issue #419, opened 2016-10-03
Filter checkbox not being checked for place


jamieu:
This might already be fixed in one of the branches, but when searching via a place the checkbox isn't selected.

/search?places=Paris

(this is when following a link from an object page)

<img width="457" alt="screen shot 2016-10-03 at 18 04 36" src="https://cloud.githubusercontent.com/assets/91365/19046124/f3024c40-8993-11e6-8aff-2872f437845b.png">

jamieu:
Fixed 👍

PR #472, opened 2016-10-11
Place link

Danwhy:
ref #419, #448

Link was incorrect for filtering by place.

Also changes the search field from 'name.value' to 'summary_title', as all items with a name have a summary value, but some items only have a summary value, with no name. This ensures we are displaying all the possible filters (ref #449)


nelsonic:
@Danwhy simple enough. (_looks good_) 👍

Issue #473, opened 2016-10-11
Image for document not displaying on SERP page


Danwhy:
related to #432 
Images are showing on document page if they exist, but not on the search results

jamieu:
Waiting on images to test

jamieu:
Looking good.

PR #476, opened 2016-10-12
displays document images on result page

Danwhy:
ref #473


nelsonic:
Yeah, looks good. merging. 👍

Issue #442, opened 2016-10-06
Displaying incorrect value for "PART OF ARCHIVE:"  on document pages


jamieu:
Arrive here: /documents/smga-documents-110000003
Visit here (via archive browser): /documents/smga-archive-110000050

The "PART OF ARCHIVE:" value should be the top level of the archive (The Babbage Papers) rather than the current pages's parent (Notations of the Analytical Engine).

<img width="865" alt="screen shot 2016-10-06 at 13 10 45" src="https://cloud.githubusercontent.com/assets/91365/19151756/6837625a-8bc6-11e6-9399-e81a3916b3c5.png">
<img width="916" alt="screen shot 2016-10-06 at 13 10 39" src="https://cloud.githubusercontent.com/assets/91365/19151757/6906fe98-8bc6-11e6-801c-d874ccc8f8fe.png">

jamieu:
👍

PR #477, opened 2016-10-12
displays correct value for 'part of archive'

Danwhy:
ref #442


nelsonic:
Gotta love those Fonds facts ... 👍 
(_looks good. merging_)

Issue #403, opened 2016-09-28
Company names being incorrectly displayed in filters as object type.


jamieu:
See multiple entries for 'Turner Bros." in filters, these should a) not be appearing under 'Object type' b) ideally (just) one of them should be appearing under people. 

My guess is the are coming form that single archive record/document and this is due to the same field name being used in the index for Mimsy and AdLib records to  represent different things (ie. .name in Mimsy is the object 'type' name, whereas in AdLib records in refers to the name of a person). 

Need to investigate index / code and see what is causing this.

<img width="812" alt="screen shot 2016-09-28 at 12 46 54" src="https://cloud.githubusercontent.com/assets/91365/18912326/b01409a0-8579-11e6-9f2d-5452cd224262.png">

SimonLab:
resolved with #406

PR #478, opened 2016-10-13
fixes org-name as type bug again

Danwhy:
#403 was reintroduced in the refactoring of the filters.

This pr fixes it and adds a regression test to make sure it doesn't happen again


nelsonic:
Dang! thanks for fixing it @Danwhy 👍

PR #487, opened 2016-10-17
Homepage

tobystokes:
There are few bits of editorial changes still to come I expect, but worth merging in now so people can see on staging


nelsonic:
@tobystokes looks great! _exciting_! 🎉

nelsonic:
Merging so we can see it on `staging`. 
can always be improved from there. 
thanks! 👍

Issue #480, opened 2016-10-13
Add museums as pillboxes to the search bar


iteles:
Once #357 is complete, we will need to add them to the pill boxes in the search bar.

jamieu:
Looks good

Issue #445, opened 2016-10-06
Add 'Archive' as a sticky filter (same as on display/category)


jamieu:
Add 'Archive' as a sticky filter (same as 'on display' and 'category')

And ensure it appears when linking though from 'search within' box on archive pages (or if a user filters by an 'archive' in the filters).

<img width="803" alt="screen shot 2016-10-06 at 13 20 10" src="https://cloud.githubusercontent.com/assets/91365/19151984/a0ce43da-8bc7-11e6-9153-033a056b47a3.png">

jamieu:
Still not seeing the pill box if I "search within" (via the archive browser on a document page) 

Doc page:
<img width="1174" alt="screen shot 2016-10-31 at 16 19 56" src="https://cloud.githubusercontent.com/assets/91365/19861823/e9bfeeb4-9f85-11e6-8498-2e6808a324a9.png">

SERP page after search:
<img width="878" alt="screen shot 2016-10-31 at 16 20 03" src="https://cloud.githubusercontent.com/assets/91365/19861824/e9c1e08e-9f85-11e6-890a-32eb09307632.png">

jamieu:
👍

Issue #438, opened 2016-10-05
Add gallery as a filter


jamieu:
Just a reminder that we need to add this (in addition to the museums) once we have location/facilities.

But only the museum should appear as pill boxes.

jamieu:
Looks good 👍

Issue #435, opened 2016-10-05
Assign museum links to the 'on display' widget


jamieu:
My thought is for now it would simply be a mapping between museum (code ie. NRM) and four maps one per museum (which we already have).

@tobystokes will need your help styling this once we have some objects that are appearing on display.

jamieu:
Just been informed that we don't have a consistent set of PDF maps to use here.

So let's just link to the relevant museum's visit page.

- http://www.sciencemuseum.org.uk/visitmuseum
- http://www.nrm.org.uk/planavisit/
- http://www.nationalmediamuseum.org.uk/planavisit
- http://msimanchester.org.uk/visit

tobystokes:
styling is #54 - but on hold until example object is viewable

jamieu:
Closing for now, might want to revisit once we have the new CMS / gallery pages.

PR #495, opened 2016-10-18
adds location (as gallery or museum) as a filter

Danwhy:
* Adds galleries and museums to filters/facets. ref #274, #438 
* Allows use of short urls ('/scm', '/nrm', '/msi', '/nmem') that redirect to searches with relevant museum filters applied. ref #357 
* Adds museum and archives as sticky filters. ref #445, #480 
* Adds museum and gallery to on display box on object page. ref #435


lib/facets/aggs-all.js
```diff
@@ -103,6 +103,22 @@ module.exports = function (queryParams) {
...
+          museum_filters: {
+            terms: {size: 25, field: 'locations.name.value'}
+          }
+        }
+      },
+      gallery: {
+        filter: filter(queryParams),
+        aggs: {
+          gallery_filters: {
+            terms: {size: 25, field: 'locations.name.value'}
```
SimonLab: Why do we need to precise the size in this case?

Danwhy: Because I'm filtering the results when displaying the facets, if I have less than this, only a few of the galleries display, (and only three of the four museums), getting 25 results back ensures I have about ten galleries to display (the default size is 10)

nelsonic:
@Danwhy code looks good. please fix merge conflicts so it can be merged. thanks! 👍

nelsonic:
@Danwhy thanks for updating. 
(@SimonLab as always, thanks for doing initial review...)
Merging. 👍

PR #505, opened 2016-10-24
fixes json appearing on use of back button 

### LINKED ISSUE #504: JSON appearing when using back button from external page
If you use the back button to come onto a page after navigating somewhere else, if the last page you visited was formed by a json request, the page will just display the raw json instead of the html.  

After a bit of investigating I found that this was a Chrome issue, where it caches the requests, and gives you the most recent one when you use the back button.

Working on a fix now

Danwhy:
fixes #504

### DIFF
--- a/client/routes/document.js
+++ b/client/routes/document.js
@@ -18,7 +18,7 @@ function load (ctx, next) {
       headers: { Accept: 'application/vnd.api+json' }
     };
     var id = ctx.params.id;
-    var url = '/documents/' + id;
+    var url = '/documents/' + id + '?ajax=true';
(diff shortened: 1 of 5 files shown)


SimonLab:
I guess you find the solution in this [post](http://stackoverflow.com/questions/9956255/chrome-displays-ajax-response-when-pressing-back-button). It might be worth to have a reminder in the readme/wiki or in the issue about the "ajax" query parameter role, I think it can bring some confusion later on if the code is being refactored.

nelsonic:
`'?ajax=true'` _indeed_... 😉 
thanks for fixing that. merging. 👍

PR #503, opened 2016-10-21
Nested galleries

Danwhy:
ref #500 
Adds logic where the checking of a gallery checks the relevant museum and unchecking of museum unchecks relevant galleries

Issue #459, opened 2016-10-06
Text for on display box


jamieu:
On display

e.g. in the on display box (if possible)

If you are visiting to see this object, please contact us [link to https://group.sciencemuseum.org.uk/about-us/collection/contact (needs creating)] in advance to make sure that it will be on display.

jamieu:
I might actually update this text pre-launch, but is exactly what I asked for, so closing!

PR #511, opened 2016-10-25
Small location fixes

Danwhy:
## Merge after #503 
ref #459, adds small text changes. Also fixes relevance boost for object with locations that changed in the index


nelsonic:
@Danwhy looks good. thanks. 👍

PR #520, opened 2016-10-26
fixes Safari display bug and normalises margins

tobystokes:
After showing and discussing with @jamieu


nelsonic:
pesky safari bug! 🐞 
code looks good. merging. 👍

Issue #522, opened 2016-10-27
Change ids item on 404 page


SimonLab:
Change the [fixtures](https://github.com/TheScienceMuseum/collectionsonline/blob/master/fixtures/404.js) of the 404 pages to match the right format of ids.

SimonLab:
resolved with #523

PR #523, opened 2016-10-27
change ids on 404

SimonLab:
see #522


nelsonic:
@SimonLab micro-focussed PR. 👍 
@Danwhy thanks for reviewing. ✅ 
need to get coverage back up with a few more tests ... 😉

Issue #515, opened 2016-10-25
Lowercase names in index / update front end to use them.


jamieu:
Can we swap over to `name.value_lower` and `terms.hierarchy.name.value_lower` to handle the term names? 

Both on display and when creating aggregation / filters.

---

*.name.value and *.summary_title now has 3 extra indexed fields they all have of the following appended to them _text, _lower, and _stem.

Please note these fields will not show in the documents but you can see them in the drop down on structured query in elasticsearch head 

_text uses the standard analyzer https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-standard-analyzer.html

_lower is a custom analyzer called lower_term which lower cases the field

_stem is a custom analyzer called smg_minimal which uses a english_minimal stemmer https://www.elastic.co/guide/en/elasticsearch/reference/2.3/analysis-stemmer-tokenfilter.html

jamieu:
Seem to be fixed now

jamieu:
Still seeing caps when clicking on the category name in the record details section of the page.
http://localhost:8000/objects/co8465613/maple-v-5-0-software-maple-v-5-0-software

http://localhost:8000/search/categories/Computing%20%26%20Data%20Processing

<img width="640" alt="screen shot 2017-07-12 at 18 14 24" src="https://user-images.githubusercontent.com/91365/28130224-22989540-672e-11e7-8c69-7d1239f9c719.png">

PR #529, opened 2016-10-27
uses lowercase values for aggs and display 

Danwhy:
ref #515


lib/transforms/json-to-html-data.js
```diff
@@ -325,13 +325,10 @@ function getObjectType (data, links) {
   var value;
   if (data.type !== 'people' && data.attributes.name) {
     value = data.attributes.name.map((el, i, arr) => {
-      var val = el.value;
```
SimonLab: We don't want to add a comma to separate the values?

Danwhy: no, the comma is added on the front end

SimonLab: For all the values? I add to keep a similar code for separate the occupations for a person

Danwhy: All of the values in the details block at the bottom, there's a helper that does it. I didn't apply it to the values at the top of the page

nelsonic:
Yeah agree. But we can merge. 👍

Issue #513, opened 2016-10-25
Errors due to new data structure of index


Danwhy:
The ids of items have completely changed format, from `smgc-object-209954` to `co209954`.
This has broken a few things where we expect the old format

New format:  
First letter:
smga -> a
smgc -> c  
Second Letter:
agent/people -> p
object/objects -> o
documents/archive -> a  
Numbers:
same id as before

so `smgc-object...` would become `co...`, `smga-agent...` would become `ap` etc

SimonLab:
resolved with #517

PR #531, opened 2016-10-27
removes old archive schema

Danwhy:
because of #513, archives were not loading from the search page. They were using a schema with the format of the old ids. This schema was for a query that we're no longer using, so I've just deleted it.


nelsonic:
@Danwhy yeah no point keeping this old-schema code. Thanks for removing it.

PR #534, opened 2016-10-27
stops keyboard always showing on ios safari 

### LINKED ISSUE #510: Keyboard popping up (incorrectly) on iOS
Clicking any result on the SERP page causes the keyboard to pop up.

img_3194
img_3191

Danwhy:
resolves #510


Danwhy:
I'm not sure exactly _why_ it was happening. I assume the search bar was just never losing focus, because it only happens after the search bar has been selected at least once

nelsonic:
Thanks iOS... looks good @Danwhy 👍

Issue #544, opened 2016-10-28
Add label for pillbox


SimonLab:
- [x] OnDisplay
- [x] Archive

jamieu:
<img width="403" alt="missing-on-display-prefix dupe issue" src="https://cloud.githubusercontent.com/assets/91365/19803220/6737290e-9cff-11e6-89b8-055d99b328c1.png">
<img width="682" alt="subsequent-pillbox-missing-prefix" src="https://cloud.githubusercontent.com/assets/91365/19803228/75887166-9cff-11e6-8b54-ab5fab808377.png">

jamieu:
Yeah! Like it 👍

PR #555, opened 2016-10-28
add museum and archive pillbox name

SimonLab:
see #544


nelsonic:
@SimonLab looks good. 👍

Issue #554, opened 2016-10-28
Relevance


Danwhy:
Now that we have a new index, we need to check we are searching and boosting using the correct fields for the best relevance.

We've lost some relevance we had before (see #552) due to field names changing, and we have some new fields with better analyzers we can use to further improve relevance.

Issue #552, opened 2016-10-28
Weight on display items higher


jamieu:
This works fine for the "objects" tab but in the "all" tab, the archive documents come before an on display object. The "on display" object should almost always come first (unless we have a near direct match on title or accession no.).

<img width="770" alt="weight-on-display-higher-1" src="https://cloud.githubusercontent.com/assets/91365/19803653/c4b6cdf8-9d01-11e6-9063-78f140aa90b8.png">
<img width="863" alt="weight-on-display-higher-2" src="https://cloud.githubusercontent.com/assets/91365/19803652/c4b67bb4-9d01-11e6-9820-e3777a83c50c.png">

Danwhy:
This seems to be a problem with the summary_title_text field. I'm not sure exactly what the problem with it is, but some of the archives are being boosted by an excessive amount.

We can get around it for now by just using the summary_title field, or one of the others such as summary_title_stem.

It seems the summary_title_text field has a built in boost, but I'm not sure why it's only affecting archives. Needs more investigation.  
<img width="253" alt="screen shot 2016-10-28 at 12 01 30" src="https://cloud.githubusercontent.com/assets/8939909/19804450/516236ee-9d06-11e6-8845-e19d9c72f789.png">

jamieu:
Are you still waiting to deploy the interim fix? (in the interim, till we find out why the archives are posted so much). Still seeing the archives/docs take precedence (even with a 'show all' query).

<img width="1172" alt="screen shot 2016-10-31 at 15 52 36" src="https://cloud.githubusercontent.com/assets/91365/19860794/18e4765a-9f82-11e6-9f3f-4819ca149b37.png">

jamieu:
This still doesn't seem be working, Rocket is on display but still appearing half way down the SERP page?
http://collectionsonline-staging.eu-west-1.elasticbeanstalk.com/search?q=rocket

http://collectionsonline-staging.eu-west-1.elasticbeanstalk.com/objects/co8084947/stephensons-rocket-steam-locomotive

jamieu:
Looking/working much better now.

Issue #545, opened 2016-10-28
Autocomplete not displaying people/companies


jamieu:
<img width="645" alt="auto-complete-missing-names" src="https://cloud.githubusercontent.com/assets/91365/19802843/78d001e2-9cfd-11e6-93d1-2115db2df2a4.png">

SimonLab:
same issue describe [here](https://github.com/TheScienceMuseum/collectionsonline/issues/552#issuecomment-256893765)

jamieu:
Looking good 👍

PR #556, opened 2016-10-28
Query changes to fix relevance

Danwhy:
ref #552, #545

Something is wrong with the summary_title_text field, so switching it out for summary_title_stem for now. 
Need to look further into it when improving relevance (#554)


nelsonic:
@Danwhy looks good. merging. 👍

Issue #549, opened 2016-10-28
Lose "In" from "Made: in <place>" if no other values.


jamieu:
See below..

<img width="330" alt="lose - in - if we only have place" src="https://cloud.githubusercontent.com/assets/91365/19802997/59e59980-9cfe-11e6-9f7a-9e042eb5cdc9.png">

jamieu:
LGTM!

PR #557, opened 2016-10-28
display "in" if date and place are defined

SimonLab:
see #549


nelsonic:
@SimonLab looks great. 👍 
merging. 🚀

Issue #494, opened 2016-10-18
Page jumping when selecting a filter or tab


jamieu:
I'm noticing quite a bit of jumping when I select filters or tabs.
Anything we could do to reduce this would be really good.

SimonLab:
This might be resolved with #474

iteles:
Not yet, requires more investigation.

Danwhy:
This appears to just be a safari issue. 

We can resolve it by adding a setTimeout of about 200 milliseconds around the re-rendering of the page (only if the user-agent is safari), but it feels a little hacky.

We could also resolve this by only re-rendering the lower parts of the search page, where the data actually changes, but this would take a little longer to implement.

@jamieu @nelsonic @SimonLab @iteles Any ideas/preferences?

jamieu:
Maybe take a two pronged approach,

Use the setTineout for now but revisit to just display bottom half / the
results in a final pre-launch sprint.

Danwhy:
It was actually simpler than I thought to just render the changed parts of the search page, so I went ahead and implemented that #565

jamieu:
Ah fantastic :-)

jamieu:
Seems better already, have you actually deployed the change?

Danwhy:
Yeah, this has been deployed

jamieu:
Cool 👍 

Closing..

PR #565, opened 2016-10-31
only re-renders search results on search page

Danwhy:
If already in the search page, only the results are now re-rendered, rather than the whole page.
Solves issue #494

### DIFF
--- a/client/routes/search.js
+++ b/client/routes/search.js
@@ -62,8 +62,19 @@ function render (ctx, next) {
   window.scrollTo(0, 0);
   hideKeyboard();
   loadingBar.end();
+
+  var searchResults = document.querySelector('.results-page');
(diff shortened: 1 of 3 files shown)


SimonLab:
The code looks good to me. However it breaks at the moment on my localshot because of the fetch function and the last version of the fetch-ponyfill installed on my node_modules, but this should be resolved when merged with master.

nelsonic:
@Danwhy agree, code looks good. 
We _need_ to try and get coverage back up to 100% before we handover the project.
So that if any changes need to be made between now and "Phase Two" the changes will be "_rejected_" by CodeCov if the testing coverage decreases ...

Danwhy:
@nelsonic We should have 100% coverage again once the images are back in the index

PR #566, opened 2016-10-31
displays all places made if more than one

Danwhy:
ref #559

Issue #271, opened 2016-08-05
Related objects (on object page)


jamieu:
Ideally we'd use WHOLES/PARTS and other explicitly defined relationships between objects.

But for now:
- [ ] Objects on display in the same location (gallery)
- [ ] Objects that have the same category and object types

The first "Objects on display in the same location (gallery)" would be a quick win, the second probably needs more thought / tweaking / adjustment. Therefore let's go with "on display in the same location" first and see how that looks.

@emilyfildes does this reflect your thoughts / our previous conversations re. this?

jamieu:
- [ ] Objects on display in the same location (gallery)
- [ ] Objects that have the same category and object types

jamieu:
Much better!

PR #568, opened 2016-10-31
Similar objects

Danwhy:
### To be merged after #566 

ref #271 
Gets objects on display in the same museum or in the same category and displays them in the related objects section.


nelsonic:
@Danwhy looks good. ✅ 
@SimonLab thanks for initial review. 👍

Issue #570, opened 2016-10-31
Images for twitter/facebook cards


Danwhy:
Reminder to add images to the facebook/twitter meta tags when images are back in the index.

jamieu:
Looks good, need to make sure we test before launch (in fact let me know on Slack and we can take the auth off for an hour to test).

Here's are a couple of Twitter / OpenGraph validators
https://cards-dev.twitter.com/validator
http://validator.w3.org/#validate_by_input
http://opengraphcheck.com/
http://iframely.com/debug/

jamieu:
Cards don't seem to be validating?
http://collection.sciencemuseum.org.uk/objects/co26704/rocket-locomotive-steam-locomotive

<img width="1064" alt="screen shot 2016-12-02 at 16 03 42" src="https://cloud.githubusercontent.com/assets/91365/20840784/4e0be558-b8a9-11e6-9fea-ccde905d87a0.png">

jamieu:
Validating nicely for Twitter & OG - looks cool too :-)

But getting a `"No Character encoding declared at document level"` error from http://validator.w3.org/check.

And a 406 HTTP error from http://iframely.com/debug/ (not the end of the world - but be good to check why we're getting a [406](http://www.checkupdown.com/status/E406.html) back).

Not showstoppers, but be good to cast an eye over both of them.

Issue #509, opened 2016-10-24
Add OpenGraph metadata to pages


jamieu:
I know we've done this for [Twitter](https://github.com/TheScienceMuseum/collectionsonline/blob/32dceac95bb2a15b5f16c9e7b39f8922b3da1519/templates/layouts/default.html) and [Schema.org](https://github.com/TheScienceMuseum/collectionsonline/blob/81d510a894bda1417b249bc76efc685fa9c6886c/lib/transforms/json-ld.js) (which should cover PinInterest) but be good to add some basic OpenGraph markup for Facebook as well.

I think we'll probably just have to go with an og:type of 'article' (http://ogp.me/#types), unless anyone can find a more museum object specific tag. But would be good to fill in any of the other obvious tags we have data for (ie. people, places etc.). We can always revisit later, but be good to have the basic in for launch.

https://developers.facebook.com/docs/sharing/webmasters#markup
https://developers.pinterest.com/docs/rich-pins/articles/

```
<!-- Open Graph data -->
<meta property="og:title" content="Title Here" />
<meta property="og:type" content="article" />
<meta property="og:url" content="http://www.example.com/" />
<meta property="og:image" content="http://example.com/image.jpg" />
<meta property="og:description" content="Description Here" />
<meta property="og:site_name" content="Site Name, we should pull this from somewhere global so it matches the page title" />
```

jamieu:
Looks good!

PR #571, opened 2016-10-31
adds opengraph meta tags 

Danwhy:
ref #509
Need to remember to add images once they're back in the index -  #570


iteles:
Looks good @Danwhy, thanks for opening the issue on the images too 👍

Issue #574, opened 2016-11-01
site not working on Internet Explorer 11


Danwhy:
After some investigation it seems Object.assign and Promises were throwing errors for ie11.

As we're already using babel for some transformations, we're going to add polyfills for these methods for now, we'll look into better solutions in the future.

nelsonic:
@Danwhy 
+ https://www.npmjs.com/package/object-assign
+ https://github.com/then/promise

Prefer to be _specific_ instead of loading the _monster_ that is Babel Polyfill and force users to download a potentially _large_ bundle they won't need ... 😕

Danwhy:
I agree it will be better to be specific. There are some more problems other than those two though. We're investigating some more now...

SimonLab:
I'm currently doing more tests on a real IE 11, and it seems that document.activeElement that we use [here](https://github.com/TheScienceMuseum/collectionsonline/blob/b941963f371031e58604187a0bbd698c8d5ac65a/client/lib/hide-keyboard.js#L3) to hide the keyboard on IOS returns undefined so the function blur can't be applied and returns an error which break the application. A simple solution is to just check that the value is not undefined before applying blur

jamieu:
LGTM

PR #575, opened 2016-11-01
adds polyfills for ie11 support

Danwhy:
See #574

### DIFF
--- a/client/lib/polyfills.js
+++ b/client/lib/polyfills.js
@@ -1,0 +1,96 @@
+function promisePolyfill () {
+  if (!window.Promise) {
+    window.Promise = require('promise');
+  }
+}
+
+// Polyfills from MDN - https://developer.mozilla.org
+function removePolyfill () {
+  if (!('remove' in Element.prototype)) {
+    Element.prototype.remove = function () {
+      if (this.parentNode) {
+        this.parentNode.removeChild(this);
+      }
+    };
+  }
+}
+
+function findPolyfill () {
+  if (!Array.prototype.find) {
+    Object.defineProperty(Array.prototype, 'find', {
+      value: function(predicate) {
+       'use strict';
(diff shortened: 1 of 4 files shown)


nelsonic:
@Danwhy this _appears_ to "_solve_" the issue, however ... https://github.com/TheScienceMuseum/collectionsonline/issues/574#issuecomment-257612798

Danwhy:
@nelsonic I've swapped out the babel polyfill for all the individual polyfills we need for ie11

client/lib/polyfills.js
```diff
@@ -0,0 +1,96 @@
+function promisePolyfill () {
+  if (!window.Promise) {
+    window.Promise = require('promise');
```
nelsonic: this will _always_ require the `promise` library. _right_...?
What we need is to _conditionally_ load the library as a script from the CDN rather than bundling it.

package.json
```diff
@@ -40,6 +40,7 @@
   },
   "devDependencies": {
     "autoprefixer": "^6.3.6",
+    "babel-plugin-transform-object-assign": "^6.8.0",
```
nelsonic: prefer to _limit_ our usage of babel plugins as much as possible. 
This particular does not _appear_ to have tests: https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-object-assign ... unless I'm missing something?

nelsonic:
@Danwhy this code is moving in the right direction.
if you can _confirm_ that it's working in `IE 11` we can merge an check on staging. 
Thanks! 👍

SimonLab:
@Danwhy @nelsonic I've done more tests on IE 11 and everything looks fine now :+1:

nelsonic:
@SimonLab thanks very much for confirming it works _as expected_ in `IE11` ✅ 
merging.

Issue #491, opened 2016-10-17
uncomment test image 


SimonLab:
see #489 

need to update this [test](https://github.com/TheScienceMuseum/collectionsonline/blob/master/test/client/images.clienttest.js) and this [one](https://github.com/TheScienceMuseum/collectionsonline/blob/master/test/images.test.js) when the multimedia property is back in the index

jamieu:
Can I close this or is it still WIP / a reminder?

PR #591, opened 2016-11-21
updates get image functions for new index structure

Danwhy:
Small updates to bring the code into line with the new index fields for the images and make the one image we currently have display correctly.

We've uncommented one test now, which brings the coverage up. Still won't quite be up to 100 until we have a few more images with the different values we can test (#491)

(diff coverage is only at 80% because I updated the code for the document images too, which still don't exist)


nelsonic:
@Danwhy code looks good. Codecov is being silly. (_coverage has not dropped 17% in this PR...!_)

Issue #583, opened 2016-11-09
Phillip's not found when searching on "Phillips"


jamieu:
http://collectionsonline-staging.eu-west-1.elasticbeanstalk.com/objects/co64127/phillips-economic-computer-analog-computer

jamieu:
👍

Issue #578, opened 2016-11-02
We seem to have lost relevance / ranking


jamieu:
Puffing Billy [1862-2]
/objects/co26701/locomotive-puffing-billy-with-tender-10-fish-bellied-rails-12-chairs-6-sleepers-shovel-3-firing-irons-steam-locomotive

Is 'on display' and tagged with a category 'Locomotives and Rolling Stock' ([which is a boosted category](https://github.com/TheScienceMuseum/collectionsonline/wiki/Ordering-and-relevance)), but it's appearing at the bottom of the SERPS page. 

Any idea what has changed?

iteles:
@danwhy was talking about this on Friday. One of the things we've noticed is that the relevance has been doing some odd things as the index has changed. I would suggest it's our highest priority for when the new index comes in and we do our final sprint. 

Or do we need to allocate some days to look into this sooner @jamieu?

jamieu:
Agree, lets leave it till we have the (hopefully) final pre-launch /
launch index
in place.

jamieu:
Priority for order:

- Location
- Image

jamieu:
If I do a filer on gallery alone, I'm still seeing records with no images showing before those with records. ie. this search for objects in a gallery.
http://collectionsonline-staging.eu-west-1.elasticbeanstalk.com/search?filter%5Bmuseum%5D=Science%20Museum&filter%5Bgallery%5D=Winton%20Mathematics%20Gallery&page%5Bsize%5D=50&page%5Btype%5D=search

jamieu:
Much better - closing this ticket, but lets pick up default order (past the image and location).

PR #596, opened 2016-11-22
relevance tweaks

Danwhy:
Prioritises location, then image for results: #578
Uses best fields rather than cross fields, provides better results for terms with apostrophes: #583

### DIFF
--- a/lib/search.js
+++ b/lib/search.js
@@ -26,7 +26,7 @@ module.exports = (elastic, queryParams, next) => {
             should: [{
               multi_match: {
                 query: queryParams.q,
-                type: 'cross_fields',
+                type: 'best_fields',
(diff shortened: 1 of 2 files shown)


lib/search.js
```diff
@@ -80,7 +80,7 @@ module.exports = (elastic, queryParams, next) => {
           filter: {
             exists: {field: 'locations'}
           },
-          weight: 1.2
+          weight: 10
```
Danwhy: Yeah, location > image, but most of the objects that have images are on display anyway, so they end up doubly boosted.

SimonLab: :+1:

nelsonic:
LGTM2. 👍

Issue #499, opened 2016-10-20
Sorting of image order


jamieu:
Ensure that images (assigned to single record) are sorted by:

-          PRIORITY (low/medium/high)
-          SORT_DATE_OF_UPLOAD

Will supply exact field names once latest index is updated

_(ideally we'd also have a pagination field, but this isn't yet available)._

SimonLab:
@jamieu from our discussion this morning it seems that the priority property will be a value between 0 and 9 and from the database it looks like the sort property is a string which represent the date of the upload:
![sort-multimedia](https://cloud.githubusercontent.com/assets/6057298/20534989/226b7ab0-b0db-11e6-8a7e-803d2bb60ecb.png)
Is this right and will this format will always be the same?

jamieu:
Exactly!

jamieu:
Exactly that!

jamieu:
👍

PR #597, opened 2016-11-22
sort images by priority and date upload

SimonLab:
see #499
This PR add a function to sort the list of images by priority and sort properties.


lib/helpers/multimedia/sort-images.js
```diff
@@ -0,0 +1,28 @@
+/*
```
SimonLab: I've created this new folder to start to extract the function which are inside lib/jsonapi-response.js. I think it's a bit more readable and we can also unit test this functions. Happy to put it back to jsonapi-response file if we consider that the function only be used there, any recommendation @Danwhy?

Danwhy: Sounds like a good idea to me

nelsonic:
@SimonLab looks good. 👀 ✅ 
@Danwhy thanks for checking. 👍

Issue #601, opened 2016-11-22
'Next' (and 'Prev') button throws error


jamieu:
This is on a search for 'Locomotives' 

<img width="807" alt="screen shot 2016-11-22 at 19 42 49" src="https://cloud.githubusercontent.com/assets/91365/20539291/6e33e94e-b0ec-11e6-94e3-f67d79ff4e27.png">

<img width="1125" alt="screen shot 2016-11-22 at 19 25 32" src="https://cloud.githubusercontent.com/assets/91365/20538507/77b898f0-b0e9-11e6-8dcd-973384101d9c.png">

SimonLab:
The issue appears only with the ajax request. It's a problem with the ajas query parameter type:
![error-next-page](https://cloud.githubusercontent.com/assets/6057298/20561114/bc7dc330-b174-11e6-837d-2232bce67873.png)

The ajax value is at the moment an array with [true, true] because the url of the next page seems to have twice the ajax=true query parameter:
```/search?q=rocket&page%5Bnumber%5D=3&ajax=true&ajax=true```
We need to check if the ajax=true before [adding it to the url](https://github.com/TheScienceMuseum/collectionsonline/blob/7b7b88ac601c0d3799d677dc67aa2249ac9f2f69/client/routes/search.js#L41)

jamieu:
👍

PR #606, opened 2016-11-23
update test for ajax query parameter

SimonLab:
see #601,

This PR ensure that the query parameter ajax=true is only define once in the ajax queries otherwise joi validation failed and return an error.


nelsonic:
@SimonLab looks good. merging. 👍 
@Danwhy thanks for reviewing. ❤️

Issue #602, opened 2016-11-22
Problem displaying a single image?


jamieu:
@Danwhy I can see the thumb for this on the SERP's page and if I right click I can open the main image in a new window, but I can't see the image on the page in Safari or Chrome. 

I wonder if it's to do with there only being one image?

http://collectionsonline-staging.eu-west-1.elasticbeanstalk.com/objects/co8405990/beyer-garratt-articulated-stea-steam-locomotive

btw. we shovel be using the fixed height (576) medium image as it's resized to keep it's aspect ratio based on a height resize.

```
"medium": {
              "format": "jpeg",
              "location": "",
              "location_is_relative": true,
              "measurements": {
                "dimensions": [
                  {
                    "dimension": "height",
                    "units": "pixels",
                    "value": 576
                  },
                  {
                    "dimension": "width",
                    "units": "pixels",
                    "value": 866
                  }
                ]
              },
```

jamieu:
This looks good on staging now.

PR #607, opened 2016-11-23
single image display

Danwhy:
Fix so that if an object only has one image it displays correctly. Also removes example images and uses correct image sizes. ref: #602


nelsonic:
@Danwhy code looks good to me. 👍 
@SimonLab any thoughts/comments?

nelsonic:
Codcov is being silly. this is fine to merge. 👍

Issue #608, opened 2016-11-23
Prevent autocomplete when key is enter key


SimonLab:
The autocomplete doesn't need to be executed when the key pressed is "Enter". At the moment the autocomplete is [trigger for each keyup event](https://github.com/TheScienceMuseum/collectionsonline/blob/3ab10c698aa0bc7d54108f53c41da6cb99470021/client/lib/search-box.js#L13). This will reduce the numbers of requests for a search.

PR #609, opened 2016-11-23
do not trigger autocomplete on Enter key

SimonLab:
see #608

At the moment for each search triggered with the "enter" key an unused autocomplete request is also trigger. This PR check the code of the key before executing the autocomplete.

### DIFF
--- a/client/lib/search-box.js
+++ b/client/lib/search-box.js
@@ -12,7 +12,7 @@ module.exports = function () {
   });
   searchinput.addEventListener('keyup', debounce(function (e) {
     var q = e.target.value;
-    if (q.length > 0) {
+    if (q.length > 0 && e.code !== 'Enter') {
       const requestId = currentRequestId = Date.now();
       const url = `/autocomplete?q=${encodeURIComponent(q)}`;
       const opts = { headers: { Accept: 'application/vnd.api+json' } };
--- a/client/lib/search-box.js
+++ b/client/lib/search-box.js
@@ -12,7 +12,7 @@ module.exports = function () {
   });
   searchinput.addEventListener('keyup', debounce(function (e) {
     var q = e.target.value;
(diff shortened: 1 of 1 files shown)


nelsonic:
@SimonLab I'm _happy_ with this micro-change. 👍

client/lib/search-box.js
```diff
@@ -12,7 +12,7 @@ module.exports = function () {
   });
   searchinput.addEventListener('keyup', debounce(function (e) {
     var q = e.target.value;
-    if (q.length > 0) {
+    if (q.length > 0 && e.code !== 'Enter') {
```
SimonLab: I've checked the value linked to ```e``` and it seems that 'Enter' is a right value. I'll check if it's specific to a certain browser and if it exists a better value

Danwhy: Enter is the [correct value](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values), but maybe we should use [event.key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) instead, as it has better support than [event.code](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code)

nelsonic:
Restarted the build: https://travis-ci.org/TheScienceMuseum/collectionsonline/builds/178567744 ... ⌛️

nelsonic:
@SimonLab one test fails on the build: https://travis-ci.org/TheScienceMuseum/collectionsonline/builds/178567744#L4885

SimonLab:
Thanks @nelsonic it seems that the data from Elasticseach as slightly change, I will investigate :+1:

nelsonic:
@SimonLab thanks for confirming that. merging. 👍

Issue #586, opened 2016-11-11
We seem to be experiencing a significant number of ''Error getting data from the server" messages popping up


jamieu:
This error was occurring when searching for an accession number not in the index. 

But I have noticed it popping up more more frequently elsewhere as well.

<img width="301" alt="screen shot 2016-11-11 at 13 35 04" src="https://cloud.githubusercontent.com/assets/91365/20216656/b6241a44-a813-11e6-88f1-604516532edc.png">

SimonLab:
This is due to this error:
![error-empty-search](https://cloud.githubusercontent.com/assets/6057298/20570887/3e1d9ec4-b19d-11e6-9195-dfe800021677.png)

Which is catch by in the function [get-data](https://github.com/TheScienceMuseum/collectionsonline/blob/3ab10c698aa0bc7d54108f53c41da6cb99470021/client/lib/get-data.js#L16).

We display an "no result page" which doesn't have the same DOM elements as a result search page so adding/deleting class to elements which are not present break the page.

jamieu:
👍

PR #610, opened 2016-11-23
check existing dom elts on no result page

SimonLab:
see #586

Check if dom elements exist before manipulating them.


nelsonic:
@SimonLab looks good. 👍 
@Danwhy thanks for reviewing. 😉

Issue #582, opened 2016-11-09
Dates are behaving strangely


jamieu:
- [ ] Date field not being populated from search query url
http://collectionsonline-staging.eu-west-1.elasticbeanstalk.com/search?filter%5Bdate%5Bfrom%5D%5D=1892&page%5Bsize%5D=50&page%5Btype%5D=search
- [ ] Results of a date filter are returning results that don't match the date range (see second screenshot) - the first result has a single year which doesn't match the date range.
- [ ] Date box remains closed when a search is re-run.
- [ ] In some case the date is showing correctly for the 'from' but the 'to' is todays date see screenshot).
- [ ] Date related error message appearing

<img width="1125" alt="date-bug-nov-9th" src="https://cloud.githubusercontent.com/assets/91365/20148756/f5d72d94-a6a5-11e6-9dae-35aa1adf80df.png">

<img width="1145" alt="screen shot 2016-11-09 at 17 54 36" src="https://cloud.githubusercontent.com/assets/91365/20148670/a3c5e432-a6a5-11e6-8e22-42bc41ab145a.png">

<img width="427" alt="screen shot 2016-11-09 at 17 59 16" src="https://cloud.githubusercontent.com/assets/91365/20148808/44cbe93a-a6a6-11e6-8a5e-c65253bffb60.png">

Danwhy:
I've managed to get the date filters to return the correct results, but I had a question about the exact logic we want to use.

For objects there's no problem, there is one date, the creation, and any item that was created in the specified date range is returned.

For people though, we need to decide what results to return:
 * anyone who was born in the date range
 * anyone who died in the date range 
 * anyone born and died in the date range
 * anyone that lived for any period in the date range
 * anyone who lived through the whole date range 

And what if only one date is entered?
 * first date box:
   * Anyone that was born in that particular year
   * Anyone who was born after that year
   * Anyone who was alive in that year
 * second date box:
   * Anyone who died in that year
   * Anyone who died before that year 
   * Anyone who was alive in that year

My opinion is that we should return anyone that lived for any period between the dates (so if a user enters 1900 - 1999, anyone who was alive at any time in the 20th century would be returned). I'm not so sure about what to do with only one date though, maybe return anyone that was alive in that year. @jamieu thoughts?

jamieu:
Much better 👍 

@emilyfildes do you want to cast an eye on this before closing it?

emilyfildes:
Hello - looks  massively improved!

[...]

Issue #576, opened 2016-11-02
Date filter not opening/populating when linking though from a record page


jamieu:
- First time you link though via the date field, the date box on the SERPS page is closed and the values not populated.
/objects/co157470/clay-tobacco-pipe-paris-france-1840-1910-tobacco-pipe
/search?date[from]=1999&date[to]=1999

- If you directly link then the values are populated but the date filter remains closed.
/search?date[from]=1999&date[to]=1999

jamieu:
👍

PR #613, opened 2016-11-24
Dates

Danwhy:
ref #582, #576

* Date logic has been fixed:
  * Searching for one date brings you all objects created in that year, and all people born after (if date[from]) or that died before (if date[to). 
  * Searching for a range of dates returns all objects created in that range, and all people that were alive for at least some of the range.

* Date errors have been fixed, and you can now submit dates by pressing enter as well as leaving the input.

* Dates stay in the input boxes when the page refreshes/are in there when the page loads if they're in the url, and the filter is open if there is a date.


Danwhy:
@nelsonic This PR fixes the nightwatch tests that are failing on master and causing deploys/pr builds to fail, so if you could merge this one first, that would be great, thanks.

nelsonic:
@Danwhy Looks great! 👀  👍 
@SimonLab thanks for reviewing. ✨

PR #617, opened 2016-11-28
Removed 'user' relationship type from user relationship list

jamieu:
Small tweak to the users fixture to remove the 'user' relationships (as it was too vague in the interface).

### DIFF
--- a/fixtures/users.js
+++ b/fixtures/users.js
@@ -1,6 +1,5 @@
 module.exports = [
   'supplier',
-  'user',
   'preparer',
   'collector',
   'patentee',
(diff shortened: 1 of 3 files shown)


jamieu:
Is this failing due to to a test for the value I removed?

Danwhy:
@jamie Probably, I'll have a look and fix it

jamieu:
Sorry....I was trying to save you fixing a minor issue (and now your
debugging it).

[still will remind me what else i need to check in future, when making
changes like this...which is all good stuff]

jamieu:
@Danwhy what was it?

Danwhy:
It was actually a problem with some unrelated tests that were just failing because the index changed, nothing to do with what you did

jamie:
@danwhy wrong @jamie.

Danwhy:
Oops, sorry. Too many different usernames on different platforms

nelsonic:
@jamieu looks good. merging. 👍

Issue #611, opened 2016-11-24
Suppress cite block on people pages..


jamieu:
Can we just comment it out for now on:

- All people records
- All archive records

(we may add it back at some point/if we have longer biogs. so just comment out rather than totally remove - but as it stands with the current/average biography length it pushes the related objects/docs too far down the page).

<img width="738" alt="screen shot 2016-11-24 at 12 52 58" src="https://cloud.githubusercontent.com/assets/91365/20599274/249647a0-b245-11e6-9f31-37323245690f.png">

jamieu:
@Danwhy Still seeing this on Archive pages, are we still waiting for this to be merged?
<img width="1134" alt="screen shot 2016-12-05 at 11 00 38" src="https://cloud.githubusercontent.com/assets/91365/20882720/3602ffae-bada-11e6-8b2b-8da3a3ef8722.png">

Danwhy:
Sorry, I must have missed that you wanted it on archive records too, I think it's only been done for people. Should be very quick to do though, I'll get on it.

jamieu:
Thanks, yes I think I might have stuck in the Archive post the initial
request.

If you can get that in it would be great!! (makes the "thin" pages so much
more usable; as the 'archive browser' then move up the page and into view).

jamieu:
LGTM 👍

PR #619, opened 2016-11-28
Remove cite block from people pages

Danwhy:
as per #611 (removed, as it can't be commented out because it's a handlebar partial)


nelsonic:
@Danwhy & @SimonLab 👍

Issue #623, opened 2016-11-29
Unknown maker - appearing in the filter field


emilyfildes:
Hello,

Unknown maker is appearing as a filter in the search - I think this should be removed, @jamieu ?
![unknown_maker](https://cloud.githubusercontent.com/assets/22834431/20708361/6c78cc7e-b627-11e6-98f7-637892683584.png)

I've assigned as a bug, but feel free to update if it is an index issue instead.

Thanks,
Emily

jamieu:
Yes, agreed...

Can we get this supressed..ta!

jamieu:
👍

PR #636, opened 2016-11-30
Unknown makers

Danwhy:
ref #623

### DIFF
--- a/lib/facets/aggs-all.js
+++ b/lib/facets/aggs-all.js
@@ -21,7 +21,11 @@ module.exports = function (queryParams) {
         }
       },
       maker: {
-        filter: filter(queryParams),
+        filter: {
+          bool: {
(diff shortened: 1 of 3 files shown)


lib/facets/aggs-all.js
```diff
@@ -21,7 +21,11 @@ module.exports = function (queryParams) {
         }
       },
       maker: {
-        filter: filter(queryParams),
+        filter: {
+          bool: {
+            must: filter(queryParams).bool.must.concat({ not: { terms: {'lifecycle.creation.maker.summary_title': ['Unknown maker', 'Unknown photographer']} } })
```
SimonLab: nice solution! my first idea was to filter after the elasticsearch request

Danwhy: I did that at first, filtered it out when creating the facets, but realised this should be quicker

nelsonic:
@Danwhy looks good. thanks! 👍

Issue #641, opened 2016-12-01
Pull all 3 fields from image multimedia.source.legal block into view


iteles:
At the moment if not all three fields are filled in, nothing is displayed on the page. As a user, I want to see *any* information that is available from *any* field, even if all 3 aren't filled in.

jamieu:
👍

PR #643, opened 2016-12-01
shows caption for single images

Danwhy:
ref #641, caption wasn't displaying for single images because 'hidden' class was applied, which was only being removed by the carousel (which doesn't exist for single images)


nelsonic:
@Danwhy looks good. thanks for updating. 👍 
@SimonLab thanks for reviewing. ✅

Issue #645, opened 2016-12-01
Ensure we use the primary  identifier value in the Archives..


jamieu:
ie. the BAB one in the screenshot below..

<img width="387" alt="screen shot 2016-12-01 at 12 56 16" src="https://cloud.githubusercontent.com/assets/91365/20794726/9927e9c2-b7c5-11e6-969d-5abad5929157.png">
<img width="833" alt="screen shot 2016-12-01 at 12 55 58" src="https://cloud.githubusercontent.com/assets/91365/20794734/a330d6e0-b7c5-11e6-9678-8cf01bfe090e.png">

jamieu:
👍

PR #647, opened 2016-12-01
primary identifier

Danwhy:
Gets primary identifier for archive tree, as per #645


nelsonic:
@Danwhy + @SimonLab looks good. merging. (_thanks! hope your afternoon is going well..._)

PR #648, opened 2016-12-01
fixes json rendering error

### LINKED ISSUE #504: JSON appearing when using back button from external page

Danwhy:
fixes #504 without breaking #601

### DIFF
--- a/client/lib/to-json-url.js
+++ b/client/lib/to-json-url.js
@@ -15,5 +15,6 @@ module.exports = function (queryString) {
       qs[k] = qs[k].replace(/,/g, '\\,');
     }
   });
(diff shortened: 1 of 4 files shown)


nelsonic:
@SimonLab / @Danwhy please assign when ready. thanks! 👍

nelsonic:
@SimonLab thanks. merging. 👍

PR #649, opened 2016-12-01
deletes production-release script from travis yaml

Danwhy:
We're having trouble deploying to the staging area from travis since adding this, so removing it for now and deploying to production manually until we can work it out

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,16 +32,3 @@ deploy:
     on:
       repo: TheScienceMuseum/collectionsonline
       branch: master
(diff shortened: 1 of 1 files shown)


jamieu:
I agree with Daniel, lets pick this up later with separate buckets and a clearer head - for now lets just remove the extra prod. deploy liens from travis.yml

But for now we can deploy to production manually.

nelsonic:
@Danwhy please link to the issue where this was discussed/agree (_both in the commit message and in the PR description/comment_) thanks. 👍

Issue #653, opened 2016-12-01
Suppress zoom/expand button is no zoom is available.


jamieu:
(we might have to suppress the button completely, depending on how far we get with processing them tomorrow - give me a shout when your about to tackle and I can give a firm answer)

But in normal use the [X] expand button should only appear if there is a zoom available.

jamieu:
Looks good

PR #664, opened 2016-12-02
suppresses zoom button if no zoom exists

Danwhy:
ref #653


nelsonic:
@Danwhy looks good. thanks! 👍

Issue #569, opened 2016-10-31
Google Analytic - data layer


jamieu:
See comment at bottom for new format

iteles:
We can split this issue into 2 - a short (2 hour) research spike first so we know exactly how long it'll take to complete this issue. Very high priority prior to launch.

PR #672, opened 2016-12-02
add layer for people

SimonLab:
see #569 
Add google data layer.

### DIFF
--- a/client/routes/document.js
+++ b/client/routes/document.js
@@ -28,8 +28,10 @@ function load (ctx, next) {
       }
       var data = JSONToHTML(json);
       ctx.state.data = data;
-      // analytics
-      window.dataLayer.push(data.layer);
+      if (data.inProduction) {
+        // analytics
+        window.dataLayer.push(data.layer);
+      }
       next();
     });
   } else {
--- a/client/routes/document.js
+++ b/client/routes/document.js
@@ -28,6 +28,8 @@ function load (ctx, next) {
       }
       var data = JSONToHTML(json);
       ctx.state.data = data;
+      // analytics
+      window.dataLayer.push(data.layer);
       next();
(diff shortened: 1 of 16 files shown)


lib/transforms/search-results-to-template-data.js
```diff
@@ -19,6 +19,7 @@ module.exports = (queryParams, results, config) => {
   const clearCategoryLink = createClearStickyFilterLink(queryParams, rootUrl, /categories/);
   const clearMuseumLink = createClearStickyFilterLink(queryParams, rootUrl, /museum|gallery/);
   const clearArchiveLink = createClearStickyFilterLink(queryParams, rootUrl, /archive/);
+  const inProduction = config.NODE_ENV === 'production';
```
SimonLab: yes right, thanks for spotting that

SimonLab:
@Danwhy I'm now checking the status of the application "production" or "test" on the json functions step instead of the html step, this means that the config will be always define there, do you mind to have another look at the code, thanks.

Danwhy:
LGTM

client/routes/document.js
```diff
@@ -28,6 +28,10 @@ function load (ctx, next) {
       }
       var data = JSONToHTML(json);
       ctx.state.data = data;
+      if (data.inProduction) {
```
nelsonic: is `data.inProduction` based on `NODE_ENV` (_environment variable_) ?

nelsonic:
@SimonLab looks good. 😍 
@Danwhy thanks for init check. 👍

client/routes/document.js
SimonLab: yes, if NODE_ENV is different to "production" inProduction will be false

Issue #652, opened 2016-12-01
Can we move image filter to top of filters.


jamieu:
<img width="309" alt="screen shot 2016-12-01 at 20 03 39" src="https://cloud.githubusercontent.com/assets/91365/20810437/a9857c4e-b801-11e6-9416-43e9221589ad.png">

jamieu:
LGTM 👍

PR #677, opened 2016-12-04
move image filter on top

SimonLab:
see #652


nelsonic:
@SimonLab looks great! 😉 
@Danwhy thanks for checking. 👍

Issue #651, opened 2016-12-01
Use 'large' image for download button


jamieu:
If possible can we also force it to download rather than display in the browser (but that's not a big issue if not; not sure if we can force the Mime type in JS?).

```
"processed": {
            "large": {
```
<img width="318" alt="screen shot 2016-12-01 at 20 03 54" src="https://cloud.githubusercontent.com/assets/91365/20810410/94a7df2e-b801-11e6-8122-b8e7284ce507.png">

jamieu:
looks good

PR #681, opened 2016-12-05
Download image

Danwhy:
ref #651 

Uses large image for download, also makes download work for all images on carousel, and single images.


nelsonic:
@Danwhy looks good. thanks! 💄 
@SimonLab thanks for checking. 👍

Issue #669, opened 2016-12-02
Change 'Accession number" to "Object number" 


jamieu:
I would have changed it myself, but was a bit worried the key was also used as a class name.

<img width="381" alt="screen shot 2016-12-02 at 14 16 29" src="https://cloud.githubusercontent.com/assets/91365/20837000/efaf4cde-b899-11e6-9b2b-befba93fd1dd.png">

SimonLab:
The name "accession number" is the key of an [object](https://github.com/TheScienceMuseum/collectionsonline/blob/06e986aa8af55cd1f8b6769e7dd0831cb5fcbd7b/lib/transforms/json-to-html-data.js#L239). We might just need to change the name of the key to change the name displayed on the page but we also need to check if there are any other elements link to this object (I don't think so but to verify)

jamieu:
I don't mind the key stays as "accession number" so much [in fact for now that's far safer] but would be good to change what is shown on the page ie. the label.

But I think we're looping though the details array and using the key as the label (is that right?). If so I can see why that's maybe not as trivial as it seems.

Let's not get to hung up on this, if we can easily change the label lets do so, if not lets revisit next week (with a clear head).

jamieu:
Let's revisit next week with a clearer head!

Thanks for investigating.

jamieu:
Looks good!

PR #685, opened 2016-12-05
replaces accession number with object number

Danwhy:
ref #669


nelsonic:
Makes sense. merging. thanks! 👍

Issue #663, opened 2016-12-02
Format numbers over 999 with comma's ie. 1,699 | 10,000 |100,000


jamieu:
![unknown-5](https://cloud.githubusercontent.com/assets/91365/20833651/ab6233a4-b888-11e6-899d-17a76d642da0.png)

jamieu:
Yah! 👍

PR #687, opened 2016-12-05
formats large numbers

Danwhy:
ref #663


nelsonic:
@Danwhy LG. 👍 
@SimonLab thanks for checking. ❤️

PR #688, opened 2016-12-05
Commercial links

jamieu:
Updated some footer links until copy can be agreed.


nelsonic:
@jamieu looks good. merging. 👍

Issue #660, opened 2016-12-02
Searching within Archive / checkbox not selected


jamieu:
In to the search box above "browse archive" on this page enter "engine" - you get taken to a SERPS page with Babbage un-selected?
http://collection.sciencemuseum.org.uk/documents/aa110000176/sketch-13-general-arrangement-circular-showing-30-variables-5-barrels-with-reducing-apparatus-punch-card-mechanism-printing-apparatus-notes-introduction-of-great-bevel-wheel-18-march-1837

PR #689, opened 2016-12-05
makes archive filter work when linked from archive-browser

Danwhy:
ref #660


nelsonic:
@Danwhy looks good. thanks!

Issue #690, opened 2016-12-05
Location filter is broken 


jamieu:
When clicking on this location..

(ambersand in name?)

http://collection.sciencemuseum.org.uk/objects/co8012682/handbill-for-an-astronomy-lecture-by-mr-bartley-on-30-march-1827-handbill

```
Bad Request
ValidationError: child "query" fails because [" Culture Exhibition" is not allowed]

```
http://collection.sciencemuseum.org.uk/search?filter%5Bgallery%5D=Cosmos%20&%20Culture%20Exhibition&filter%5Bmuseum%5D=Science%20Museum

Danwhy:
This is definitely because of the ampersand. I'll look into it

PR #700, opened 2016-12-06
encodes display location value

Danwhy:
Stops error because of ampersands in gallery names: #690


nelsonic:
LGTM2. 👍 
(_thanks @Danwhy & @SimonLab_)

Issue #694, opened 2016-12-05
Dont truncate... text in blue box (with ..) 


jamieu:
The text should fill the box

<img width="286" alt="screen shot 2016-12-05 at 20 28 44" src="https://cloud.githubusercontent.com/assets/91365/20901230/78407910-bb29-11e6-962a-23788042008e.png">

PR #704, opened 2016-12-06
do not truncate figCaption

SimonLab:
see #694


nelsonic:
@SimonLab looks good.
@Danwhy thanks for checking.
❤️

Issue #561, opened 2016-10-28
Display multiple makers (of same type) on one line.


jamieu:
Display multiple makers (of same type) on one line.

http://collection.sciencemuseum.org.uk/objects/co410174/locomotive-cylinder-from-braithwaite-and-ericssons-novelty-locomotive-cylinder

http://collection.sciencemuseum.org.uk/objects/co503422/personal-computer-model-apple-i-personal-computer

<img width="815" alt="screen shot 2016-12-05 at 20 57 13" src="https://cloud.githubusercontent.com/assets/91365/20902074/72c6094c-bb2d-11e6-8768-b4a01f864eb0.png">

<img width="486" alt="screen shot 2016-10-28 at 16 15 07" src="https://cloud.githubusercontent.com/assets/91365/19812170/68e149ec-9d2c-11e6-9ed8-584a5791a33d.png">

jamieu:
Nice, like the use of natural language (and) too 👍

PR #709, opened 2016-12-07
Inline makers

Danwhy:
displays all makers of the same type together #561


client/lib/polyfills.js
```diff
@@ -88,9 +88,41 @@ function fillPolyfill () {
...
+      value: function(predicate) {
+        'use strict';
+        if (this == null) {
+          throw new TypeError('Array.prototype.findIndex called on null or undefined');
+        }
+        if (typeof predicate !== 'function') {
+          throw new TypeError('predicate must be a function');
+        }
+        var list = Object(this);
+        var length = list.length >>> 0;
```
SimonLab: I'm not sure I fully understand how the length value is build with the operator ">>>" From the [doc](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Bitwise_Operators#Unsigned_right_shift) I understand that it will never a negative value, is it use for that on this case?

nelsonic: I don't _understand_ why the result of calling the `.length` method would _not_ be a Number, but I'm not _offended_ by this syntax.

Issue #706, opened 2016-12-06
Only show 'notes' and 'brief biog' fields for People.


jamieu:
http://collection.sciencemuseum.org.uk/api/people/cp2736

The naming on these is confusing I know, but we shouldn't be displaying the 'biography' field despite the name.

1) This should be the 'biography' field we display if available:

```
"note": [
        {
          "primary": true,
          "value": "1823 - founded by George Stephenson with his son Robert, ......"
        }
      ],
```

2) If we **don't have a `note` field** we should display the `"type": "brief biography",`

```
{
          "type": "brief biography",
          "value": "1823-1899, locomotive manufacturer, Newcastle-upon-Tyne, 
          England, British"
        }
```

We should **never display** `"type": "biography",`

SimonLab:
@jamieu do you want to add a special field for the biography (called "biography"?) which will contain the right value extracted from "note" or "brief bio"  or do you want to still want to keep note and description (without the type briograpy) displayed? I think the description field might contain more information:
- brief biography
- web description
- biography (which will be deleted)
...
So it might be worth to keep displaying the description field maybe.

jamieu:
Hi Somin,

Not sure if this is the answer to your question (or did you mean the API vs HTML?)

We just want one biography displayed, pulled from (in order).

- "note": (if available)
- "type": "brief biography", (if available, but not if we have "note:" above)
- blank (anything else)

We **never** want to display "type": "biography" in the HTML page. The reason we didn't normalised them down to a single value in the index is that we might want to access the data in the "type": "biography" later on / debugging. The Reason the names make no sense (ie. biog==notes and notes==biog is because that is how it is in the backend collection systems and we wanted to keep it consistent; in hindsight that might not have been the best idea but...). One day I will unpick all this, but now wasn't the time :-)

jamieu:
Just to clarify this is only on the Mimsy derived people pages.

SimonLab:
@jamieu thanks for the detail. Sorry I was thinking you were referring to the api page. At the moment the api page contains a bit more information displayed than the html view.
For the html view of the people we are already filtering the value for the biography. We are taking the notes if they exist otherwise we are taking the type="web description" from the description field, so we are excluding already the "biography" type. However I've just check and it seems than the "web description" type doesn't exist anymore in the index, I couldn't find any items with the search query:
```
{"query":{"constant_score":{"filter":{"term":{"description.type":"web description"}}}}}
```
[...]

PR #712, opened 2016-12-07
Biography description

SimonLab:
refactor code while search for bug discribe on #706  (the bug wasn't one so the issue is already close now)

### DIFF
--- a/lib/get-primary-value.js
+++ b/lib/get-primary-value.js
@@ -1,4 +1,8 @@
+'use strict';
+
 module.exports = function getPrimaryValue (array) {
+  var primary;
+
   if (Object.prototype.toString.call(array) === '[object String]') {
     return array;
   }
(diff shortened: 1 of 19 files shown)


lib/get-primary-value.js
```diff
@@ -7,6 +11,11 @@ module.exports = function getPrimaryValue (array) {
     return null;
   }
 
-  const primary = array.find((d) => d.primary);
```
SimonLab: I try to avoid array.find as this file is used on the front end and I don't think the find method was supported on every browser

Danwhy: It's not, but we have a [polyfill](https://github.com/TheScienceMuseum/collectionsonline/blob/master/client/lib/polyfills.js#L18)

SimonLab:
This PR focus on refactoring the get description functions. Please don't hesitate to discard this PR if you think it doesn't add enough value.

Danwhy:
LGTM. If anything I would quite like to take this a step further, and have the 'json-to-html' file (and others) as its own directory with an index.js as the entrypoint, and all of its dependent files in there. I feel like this would make it easier to see what was happening in each directory and what files are used where.

Issue #705, opened 2016-12-06
Suppress link from Mimsy people


jamieu:
So we display them, but don't hyperlink them.

(much like we do with Nationality in the screenshot below)

(_currently these values aren't coming from a taxonomy and so only link back to other people records, they are also very unwieldy as links/filters)_.

<img width="943" alt="screen shot 2016-12-06 at 12 48 02" src="https://cloud.githubusercontent.com/assets/91365/20926047/4950291e-bbb2-11e6-8859-b00cf6ee621c.png">

SimonLab:
Hi @jamieu , just to clarify which links we should disabled, is it all the links on the top (industry, nationality, based), the related people links on the right side panel or all of them?

jamieu:
Just the place (London, Paris etc.)

At the top of the page.

jamieu:
Just the place (ie. London, Paris etc.)

PR #713, opened 2016-12-07
Suppress mimsy link

SimonLab:
see #705


nelsonic:
@SimonLab looks good. 👍 
@Danwhy thanks for checking. ✅

Issue #668, opened 2016-12-02
Revisit mapping /rest/style/paths to query string style paths  


jamieu:
One for post launch, but would like to formulate a plan for this.

ie. `/search/documents?filter[archive]=The%20Babbage%20Papers` becomes `/search/archive/The%20Babbage%20Papers`

and `/search?q=computers&filter%5Bcategories%5D=Computing%20%26%20Data%20Processing&page%5Bsize%5D=50&page%5Btype%5D=search` becomes `/search/categories/Computing%20%26%20Data%20Processing`

etc.

Similar for other filters such as images, categories, locations etc.

Danwhy:
How would this work if more than one pillbox filter (category, museum) was selected?

jamieu:
My thinking would be something like:

/category/art+railways
/category/art,railways

or

/category/art/museum/scm

Does that make sense?

Happy to work out the list of paths/routes - but would love your input on
that. Do you want me to make up some examples (and you can tell me if you
think it won't work or suggest alternatives).

I think it's just the key filters, mainly so that we can i) cut/paste
'tidy' urls ii) create 'hackable' / readable urls for the key filters iii)
allow us to put them in the sitemap as 'landing pages' that Google can then
pickup.

Danwhy:
Ok, yeah that makes sense. I'll just have to have a think about how to parse the urls on the backend, but I should be able to make it work

jamieu:
Cool,

Thinking about it the hierarchy is probably something like:

- museum
- gallery
- category

and maybe an image flag ie. /images (at the end of the path)

Danwhy:
I've managed to parse the urls on the backend, and for now I'm taking the values and plugging them into the query we're currently working with. This means that typing in/linking from the restful style url works, but it still shows in the url bar as the old style with the querystring.

Actually making the restful style urls the canonical urls would require quite a lot of refactoring on both the front and back end, so is this ok for now?

jamieu:
Sounds good, lets tackle this one in stages!!

What you've done so far sound like it's a step in the right direction,
doesn't break any existing functionality and will be easy to build upon.

Also gives us a chance to test the URL's before we make them the default
url's used on the site & appearing in the browser bar.

Do update a page in the Wiki the paths/format you went with.





Jamie

jamieu:
Looking good, just a reminder that the paths/url should all be lowercased.

(_hope that doesn't cause issues_)

Danwhy:
[...]

PR #715, opened 2016-12-08
redirects rest-style urls to query string based urls

Danwhy:
#668

### DIFF
--- a/lib/helpers/utils.js
+++ b/lib/helpers/utils.js
@@ -1,0 +1,6 @@
+module.exports = {
+  uppercaseFirstChar: function (str) {
+    if (!str) return false;
+    return decodeURIComponent(str).split(' ').map(e => e[0].toUpperCase() + e.substring(1)).join(' ');
+  }
+};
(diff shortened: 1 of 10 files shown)


lib/helpers/utils.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
nelsonic: Do we _need_ to do a `.map .join` for this method?
there appear to be a _few_ other ways of doing this: http://stackoverflow.com/questions/1026069/how-do-i-make-the-first-letter-of-a-string-uppercase-in-javascript including a CSS-only solution... 😉

Danwhy: I'm mapping and joining because I want every word in the string to be capitalised, which I don't think any of those methods do. 

CSS only would be nice, but my main use for this function is to make the url match the database entries exactly.

nelsonic: Got it. 👍

nelsonic:
Merging. ✅ 
(_great work @Danwhy_)

Issue #716, opened 2016-12-09
Images being squashed on mobile


jamieu:
Assume the image/carousel height is too high on mobile viewports?

![img_1369](https://cloud.githubusercontent.com/assets/91365/21044152/5c0eba6a-bdf2-11e6-8244-f32fc73bd39d.PNG)

jamieu:
btw. if this is an easy fix then it's very high priority.

tobystokes:
Assuming you still want images to fit the avaialble space (because if not, we'll end up with a gap somewhere) these naturally more landscape images will end up being cropped slightly to fill the space.  
if this is amenable (and i think it's generally the best fix, then add)
```
.carousel__image {
	object-fit: cover;
}
```

n.b. not supported on IE, but at this screen size, its all iOS or android.

Danwhy:
That seems to work quite nicely. I think if we also reduce the height of the carousel on mobile screens, that should reduce the amount of the image that gets cropped out (and will also help with #710 by making it easier to scroll)

tobystokes:
Yes, maybe the carousel images  should have a `max-height: 75vh`
i.e. never more than 75% of the viewport. (that's an untested suggestion though)

jamieu:
This is working great on pages with multiple images. But not those pages with a singing images.

i.e.
http://collection.sciencemuseum.org.uk/objects/co163555/carbolic-steam-spray-used-by-joseph-lister-england-1866-1870-carbolic-spray

tobystokes:
For a single image, where you probably don't want to crop, i'd suggest replacing
`height: 36rem`with `max-height: 36rem`

jamieu:
Thanks Toby!

Your design is looking - even more - fantastic now the images are stating
to load :-)

Issue #657, opened 2016-12-02
Reduce font size for copyright line on mobile (or both?)


jamieu:
So that it doesn't wrap..

@emilyfildes / @tobystokes any thoughts on this?

![db3f4be2-b884-11e6-9525-8e2859d7ea24](https://cloud.githubusercontent.com/assets/91365/20971868/f57bd520-bc8a-11e6-874f-11a8f97d45de.png)

emilyfildes:
Not crucial - if there is a font-size option that doesn't make it look like a mistake then can go for that. Or a hide / show on mobile so the whole thing is cleaner?

jamieu:
I think just reduce the font on mobile-sized view-ports.

tobystokes:
you'll need to separate out the copyright element from the main caption, but yes agree.
somethign like 
```
@include breakpoint(small only) {  font-size: 75% } //  = 12px
```

tobystokes:
but it's wrapping because there are some hard `<br>`'s in there. That may be coming from the data...

Danwhy:
I think those are just to separate the three lines - caption, copyright and usage

tobystokes:
Ah right.

(A longer term improvement might be to move them both to the Use this image menu if that would be allowed)

jamieu:
Closing for now, let's revisit once the zoom button is back

PR #720, opened 2016-12-09
Mobile images

Danwhy:
Makes wide images look better on mobile (#716) and reduces font size of caption (#657)


nelsonic:
LGTM2. 👍

Issue #654, opened 2016-12-02
Update to Dependencies to latest Hapi & Joi ?


nelsonic:
@jamieu a new version of Hapi and Joi have recently been released. From my reading of the release notes there should be a _smooth_ upgrade path for us. We will _have_ to update at some point, the question is:

> Should we consider doing this _before_ Monday's "soft launch"?

see: https://david-dm.org/TheScienceMuseum/collectionsonline

jamieu:
Agree, that looks like a pretty smooth path and given next week we're still beta and very actively monitoring/testing the site it might actually be a good time to spot any issues with the upgrade. 

Assume we can back out of we noticed it really broke something major.

SimonLab:
The only small update we need to do are linked to joi where joi.bolean and joi.date.format have change

PR #721, opened 2016-12-09
Dependencies update

SimonLab:
see #654


schemas/date.js
```diff
@@ -1,7 +0,0 @@
-const Joi = require('joi');
```
SimonLab: I don't think we use this schema anymore is that right?

Issue #739, opened 2017-01-05
Increasing Tests for Reliability


nelsonic:
Hi @jamieu,
We already have a fairly _decent_ level of testing in the application [![Coverage Status](https://img.shields.io/codecov/c/github/TheScienceMuseum/collectionsonline.svg?maxAge=2592000)](https://codecov.io/gh/TheScienceMuseum/collectionsonline)
Would it be OK to invest a couple of hours of Dev time at the start of the `next` sprint adding a few _more_ tests to get us up to ***100% Coverage***? (_please/thanks! and happy new year!_)

jamieu:
Happy new year!

Yes, almost certainly - currently planning priorities (and budgeting) for the our next steps. There's a few things I know we'll have to look at / really would like us to look at first. But expect we can find a few hours to up the test coverage :-)

jamieu:
I've had to remove a couple of tests (that were failing due to new index) to deploy an serious/emergency fix for the the internal 'notes' field displaying on objects records.

Have commented the relevant files so we can add them back later:
- test/client/end-to-end.clienttest.js
- test/client/multiple-filters.clienttest.js
- test/client/routes/document.clienttest.js

PR #758, opened 2017-01-19
update tests for new index and relevance

SimonLab:
see #739


SimonLab:
Node security is failing due to the unicode and slug package. We can try to fork these packages to resolve the security issue, @nelsonic @Danwhy any thoughts?

Issue #733, opened 2016-12-17
Errors finding assets 


jamieu:
Might be nothing to worry about, but noticing a number of broken paths in the logs - although these assets exist so not sure what is causing it.

plublic.js: no file found here:  assets/img/global/logo-scm.svg
plublic.js: no file found here:  assets/img/global/logo-smg.svg
plublic.js: no file found here:  assets/icons/symbol/svg/sprite.symbol.svg
plublic.js: no file found here:  assets/fonts/sm-din-black.woff2
plublic.js: no file found here:  assets/fonts/sm-din-regular.woff2

Seems be be [getting output here](https://github.com/TheScienceMuseum/collectionsonline/blob/master/routes/public.js) (typos come in useful sometimes!).

```
module.exports = () => ({
  method: 'GET',
  path: '/{path*}',
  config: {
    handler: function (req, reply) {
      console.log('plublic.js: no file found here: ', req.params.path);
      reply.file(Path.resolve(__dirname, '..', 'public', req.params.path));
    }
  }
});
```

SimonLab:
@jamieu you're right the consle.log will display the message even when there is a file, I think this console.log is an old debug tool that we forgot to delete. Thanks for spotting that.

PR #760, opened 2017-01-19
Clean public.js

SimonLab:
see #733

### DIFF
--- a/routes/public.js
+++ b/routes/public.js
@@ -5,7 +5,6 @@ module.exports = () => ({
   path: '/{path*}',
   config: {
     handler: function (req, reply) {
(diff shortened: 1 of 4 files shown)


nelsonic:
@SimonLab did you see `Node Security` is _unhappy_... 🤔  ❓

SimonLab:
@nelsonic yes unfortunately the unicode package used in the slug package download some data over http which doesn't seems secure: "Unicode downloads binary resources over HTTP, which leaves it vulnerable to MITM attacks." MITM is for [Man In The Middle](https://en.wikipedia.org/wiki/Man-in-the-middle_attack) I'll create an issue to see if we can replace or adapt the slug package

Issue #773, opened 2017-01-26
Missing all but first occupation (people)


jamieu:
We seem to have lost the 

- ES Index (cp36993):
```
"occupation": [
"computer pioneer", "mathematician"
],
```
vs

- JSON / Page (we're only showing the first value / we are also only showing one value in the JSON?)
http://collection.sciencemuseum.org.uk/people/cp36993

```
      "occupation": [
        "computer pioneer"
      ],

```

SimonLab:
link to #532

Issue #518, opened 2016-10-26
Clean occupation data


SimonLab:
```
"occupation": [
" computer pioneer",
"mathematician",
"mathematician; computer pioneer"
],
```
- [ ] Delete the last value
- [ ] Delete unnecessary spaces in values

jamieu:
This should now be correct in the index, @SimonLab do we need to revert anything?

SimonLab:
@jamieu yes we might delete some small changes but it should be quick, I'm going to have a look now

SimonLab:
@jamieu it doesn't seem that the changes have been applied to the index yet. I've just check on ada (agent/cp38764/) and the spaces and last values is still here:
![occupation-people](https://cloud.githubusercontent.com/assets/6057298/20679180/bd6463b4-b591-11e6-9f55-6ca88cba7f3a.png)

SimonLab:
This is now resolved in the index, see #773

PR #774, opened 2017-01-27
retreive all occupations

SimonLab:
## To be merged after #776 
The database has the now the right values for the occupations (#518) so this PR delete a fix code we used to extract the right values from the Elasticsearch index

see #773


SimonLab:
@nelsonic I'm curently working on a fix for node security issue: #775

SimonLab:
The node security issue is now resolved with #776

Issue #780, opened 2017-02-03
New route for returning X random objects within a set criteria


jamieu:
This has been something asked for re. the HackDay, it's not a high priority task but seems fairly trivial to implement and would be a great addition to have in place.

Basically, can we add an optional `random=1 ` (or better still `random=n`) param to the query params, that will return you either 1 or the number of records specified by `n` - based on the other filer params 

https://collection.sciencemuseum.org.uk/search?filter%5Bmuseum%5D=Science%20Museum&filter%5Bgallery%5D=Information%20Age%20Gallery%3A%20Cable&page%5Bsize%5D=50&page%5Btype%5D=search

ie. appending `random=3` to the end of the above would get you 3 random objects, with images in the `Information Age Gallery`. 

Would only be needed in the JSON view, but no harm having it in the HTML as well  if that simplifies things.

PR #787, opened 2017-02-06
Get random search results 

Danwhy:
ref #780

Allows addition of `random=n` as a query string, where `n` is a number between 1 and 50, and returns `n` random results

### DIFF
--- a/lib/search.js
+++ b/lib/search.js
@@ -14,7 +14,8 @@ const searchWeights = require('./search-weights');
 module.exports = (elastic, queryParams, next) => {
   const pageNumber = queryParams.pageNumber;
   const pageSize = queryParams.pageSize;
-  const body = {
+  var body = {
+    size: pageSize,
     query: {
       function_score: {
         query: {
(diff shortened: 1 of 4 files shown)


lib/transforms/search-results-to-jsonapi.js
```diff
@@ -27,7 +27,7 @@ module.exports = (queryParams, results, config) => {
 // Create resources from hits, filter any unknown resource types
 function createResources (results, rootUrl, mediaPath) {
   const hits = (results.hits || {}).hits || [];
-
+  console.log(hits.length);
```
nelsonic: This is fine to _test_ the output but will be quite annoying later.
please make a note to remove it. and/or use "Good" for Logging. see: https://github.com/dwyl/learn-hapi#logging-with-good

lib/search.js
```diff
@@ -86,14 +87,30 @@ module.exports = (elastic, queryParams, next) => {
...
     from: pageNumber * pageSize,
-    size: pageSize,
     body: body
   };
 
   if (!queryParams.q) {
     searchOpts.body.query.function_score.query = {match_all: {}};
   }
 
+  if (queryParams.random) {
```
nelsonic: a comment linking to the issue/story where this functionality was requested would be good here `else` the "_next person_" will have to _waste time_ asking: "_**Why** is this code here...?_"

test/random.test.js
```diff
@@ -0,0 +1,70 @@
+const QueryString = require('querystring');
+const testWithServer = require('./helpers/test-with-server');
+const dir = __dirname.split('/')[__dirname.split('/').length - 1];
+const file = dir + __filename.replace(__dirname, '') + ' > ';
+
+testWithServer(file + 'Request for 4 random items', {}, (t, ctx) => {
+  t.plan(3);
```
nelsonic: _technically_ you don't need `t.plan(3) if you are using a `t.end()` ...

Issue #675, opened 2016-12-02
Add pillbox for Image filter


jamieu:
in the style of the others? 

(purple/green? not sure what we're using at the moment)

PR #789, opened 2017-02-07
add images pillbox

SimonLab:
see #675


Danwhy:
The filter doesn't stay when we do a new search. I think you'll need to add it into https://github.com/TheScienceMuseum/collectionsonline/blob/master/client/lib/search-listener.js

PR #798, opened 2017-02-08
REST style urls

Danwhy:
ref #668 

@jamieu This should all be fully functional now, so if you want to test it and make sure it's all working you can pull down this branch. I just need to change a lot of the tests before it will be ready for merging


Danwhy:
@jamieu @nelsonic This is ready to be merged if you're happy with it (nsp seems to be stuck on pending)

Full documentation coming in https://github.com/TheScienceMuseum/collectionsonline-api

jamieu:
Perfect  :-)

Thanks @danwhy

I think I'm gonna wait till after the HackDay to deploy and have also
discussed getting the index mappings updated so that the categories and
other facet values have a lowercase version, which can also be deployed
post the HackDay - would we need a significant code change to utilise those
lowercase values?. Be much nicer if the urls we're case insensitive.

Danwhy:
The urls in this pr work case insensitively, the uppercased categories and other filters all seem to be consistent in that they have the first letter of each word uppercased, so I was able to use that to filter by the correct value.

Although they work, they don't show up lowercase in the url bar though at the moment.

Did you want the documentation/examples to be using the old style urls for now then?

jamieu:
MMMmmm.... documentation/examples

I think given we have ~100 people turning up tomorrow and Wednesday to use
the API and many of them have already started coding against it using the
old style urls/query params  it might be best to keep with the old format
for the next couple of days. Just a little worried about deploying anything
that will cause things to break.

Is the old format backwards compatible / will it still work after we deploy
the new code?

To save any confusion it might be best to hold off for now, save an hour
and then use it to get the new routes deployed and any docs updated post
that?

Danwhy:
It should be backwards compatible, but yeah, probably best not to risk it for now.

jamieu:
Hi @Nelsonic

We were going to encourage people to maintain their own repos (on GitHub).
Think we'll hold off on creating an org this time round (but that's a nice
idea).

Well link off to the best projects from the "About our API page" and will
be writing several blog and Medium posts about the day and projects.

Agree, be interesting to see what people use / wish they had and adjust the
examples afterwards accordingly.

jamieu:
Hi @Danwhy ,

I've checkout this branch but still seem to be seeing the old style urls `[see below]`?

I've checked the code against the PR and it looks like I am running the correct branch (`urls`) locally? `[see below]`

Any ideas - or would you expect the urls to still be using to old style paths?

Thanks

Jamie

screen shot 2017-03-06 at 13 18 54
screen shot 2017-03-06 at 13 17 37

Danwhy:
It should be working:
screen shot 2017-03-08 at 10 07 56
Are you sure the branch has been built correctly? (run it with `npm run watch`)

I would like to add some more tests before we merge this in though, just to make sure we've got full backwards compatibility

nelsonic:
@jamieu can this be re-visited now that the hackathon is over? 🤔

jamieu:
Hi @nelsonic,

Thanks for reviewing these PR's.

It looks good from my brief testing.

The index should now have been updated so that the paths/values can be checked case insensitively.

So just need to:

- Update paths to use lowercase (@danwhy - is this easy to do? Happy to make the changes myself if trivial)
- Quick 'sense check' to see once and for all if we can (eaisly'ish) use - (dashes) in places of spaces `%20` in the paths. ie check for both versions (`%20` and `-`)in the index (although we may have categories with dashes and some with spaces, we won't need to differentiate between them ie. we don't have two similar categories/museums that only differ by that one character_). Reason for asking for this now/before we deploy is i) A dash `-` look so much nicer/cleaner than `%20` in paths and more importantly ii) so that Google doent' end up indexing two sets of paths (if we change it in future).

Post merging/deploying this PR, we need to deploy changes to/from the sitemap.xml repo (I think @SimonLab has made the necessary changes there. Is that correct?). Let me know what I need to do there. Happy to pick this one up later, once we have this PR deployed and you guys have some breathing space.

jamieu:
@danwhy have noticed an issue with the pagination (https://github.com/TheScienceMuseum/collectionsonline/issues/868).

I have reverted the live site back to an old branch and have taken a look at your PR https://github.com/TheScienceMuseum/collectionsonline/pull/798/files#diff-d53192600b29fee3e0a0f94b37b15367

But I can’t spot an obvious error. Any chance you could suggest / point me in the direction of where I should be looking? 

Happy to take a look at fixing it myself, just don’t want to go hacking around too much if it’s obvious to you what might have got overlooked / what needs adding.

Issue #762, opened 2017-01-25
Zooms / complete integration


jamieu:
We now have zooms in the index / S3 - be good to get this tied up (the fronted OSD code should all be there ready to go, might just need some tweaking).

Might need to update the IP address of the zoom server to the one below.

`http://54.171.226.99/iipsrv/iipsrv.fcgi?DEEPZOOM=/s3-zooms/I/A/A/1972_36_Neg_102.ptif.dzi
`

Also:
- See if we can update the carousel code so that clicking an image opens it in zoom mode https://github.com/TheScienceMuseum/collectionsonline/issues/763
- Be good to check the status of this zoom/osd issue Daniel spotted previously: https://github.com/TheScienceMuseum/collectionsonline/issues/479
- Can we add a rotate icon in full screen mode  (let me know if we need an icon) https://github.com/TheScienceMuseum/collectionsonline/issues/742

```
            "zoom": {
              "format": "pyramid tiff",
              "location": "http://54.229.248.172/iipsrv/iipsrv.fcgi?DEEPZOOM=/s3-zooms/zoom/S/O/O/1993_8049.ptif",
              "location_is_relative": true,
              "measurements": {
                "dimensions": [
                  {
                    "dimension": "height",
                    "units": "pixels",
                    "value": 3622
                  },
                  {
                    "dimension": "width",
                    "units": "pixels",
                    "value": 4930
                  }
                ]
              },
```

SimonLab:
@jamieu I'm currently trying to access the zoom images for "Babbage's Analytical Engine" (http://collection.sciencemuseum.org.uk/objects/co62245/babbages-analytical-engine-1834-1871-trial-model-analytical-engines).
The index has the zoom properties for this document:
```
"zoom": {
"format": "pyramid tiff",
"location": "W/P/A/1878_0003__0005_.ptif",
"location_is_relative": true,
...
```
But when I try to access the image with the new IP address:
```
http://54.171.226.99/iipsrv/iipsrv.fcgi?DEEPZOOM=/s3-zooms/W/P/A/1878_0003__0005_.ptif
```

I have the following message error: ```Asked for non-existent tile: 1881```

So I was wondering if the way I built the url for the zoom is right
http://54.171.226.99/iipsrv/iipsrv.fcgi?DEEPZOOM=/s3-zooms/ + {location zoom defined in Elasticsearch}

or did I miss something on the path of the url?

jamieu:
Ah does this one work better?
[...]

PR #803, opened 2017-02-10
fixes osd crashing safari bug

Danwhy:
ref #762


jamieu:
Still seems to be happening for me on Safari 10.0.3

PR #834, opened 2017-03-21
Project springclean (Media Museum Relaunch) 

jamieu:
This PR addresses various changes to the site/codebase to reflect the new naming of the Media Museum (_now National Science and Media Museum_).

Unfortunately the index still contains the old museum names for onDisplay locations (and will continue to for some time). So I have tried to update the code so that the name of the museum is only updated at the point of display - leaving the underlying filters as they are (so they match up with the values in the index; esp. as the old/new names are still similar).

Their may (_I expect almost certainly_) be a nicer way to handle this (maybe we could add a layer directly above the index to handle the remapping?) and happy to look at it in a future sprint or whenever we all next have a chance to sit down and do some refactoring. 

But hopefully nothing I have done is too irreversible or overly icky (_although do let me know if your recoil in horror at anything!_).

I also had to update one of the fixtures `fixtures/elastic-responses/example-get-response-person.json` to reflect the correct biography field now appearing in the index and another change to `test/fixtures/copy-data.js` to reflect a user that (correctly) no longer exists in the index.


jamieu:
Thanks @nelsonic 👍

Issue #820, opened 2017-03-06
Display related articles/blog posts on RHS of related articles


jamieu:
The new sites (for each museum) will soon provide a JSON feed of all pages/articles within the main websites that contain/link to objects/people/documents.

There will be one feed per site (_currently we only have NMM_):
https://www.scienceandmediamuseum.org.uk/collection-media/collection-usage/objects

- parse feed for current page [object|person|doc] asynchronously
- cache JSON / results (~every 6 hours)
- style up the results in HTML as a RHS 'related nav' block
- should be broken down by museum (via a heading ie. "Related articles on National Science and Media Museum")
- should be broken down (sub-heading) by type ie. "Objects and Stories" or "Blog Posts" (if multiple results in different categories). _I need to get type added, so may not be able to add this in just yet, but code should be easily adapted to handle it when we do._
 
_(would be much nicer if Google could parse/follow the links back to the articles on the main sites for SEO reasons; so should think how implementation will affect this)._

PR #863, opened 2017-04-19
Articles

Danwhy:
Fetches related articles asynchronously (#820), and renders them on object pages


routes/articles.js
```diff
@@ -0,0 +1,20 @@
+var request = require('request');
+
+module.exports = () => ({
+  method: 'GET',
+  path: '/articles/{id}',
+  config: {
+    handler: function (req, reply) {
+      request('https://www.scienceandmediamuseum.org.uk/collection-media/collection-usage/objects', (err, res, body) => {
+        if (err) {
+          console.error(err);
```
SimonLab: Not sure if we need to display the error on the console

templates/pages/object.html
```diff
@@ -29,6 +29,7 @@
 
         {{#if related.people}}{{> records/record-related-people }}{{/if}}
         {{#if related.organisations}}{{> records/record-related-organisations }}{{/if}}
+        <div id="articles"></div>
 
         {{!-- {{> records/panel-taxonomy}} moved lower to record-details. but panels could also be used here. --}}
```
SimonLab: Doesn't belong to this PR but maybe we can delete this commented line in the future?

Issue #722, opened 2016-12-12
Back/Home button


jamieu:
[@danwhy / @simonlab - first, and I think only step here for you, is to set a url & maybe a `class` we can use to go back to the SERP page when appropriate. Will pick up the design side with @tobystokes separately]

This is primarily for the record pages (obj/doc/person) so that when a user runs a search and clicks though to a single record they have a means (other than the browser back button) of going back to the results set they were on.

I guess the simplest would be to set a cookie/session var on the SERPs page, so that:

- a user arriving from a SERP page would see a back button that took them back to the SERP page.
- a user arriving from Google or direct link would get a HOME button (back to home page) _or nothing_.
- the SERP page itself would always show the home button (back to home page) _or nothing_.

One gotcha might be to ensure that a user **ONLY sees the back button if coming directly from a SERP page**. We wouldn't want the back button to appear if a user goes off to another page and sees a back button which then leads them to an old/stale result set.

**1) User does a search**
<img width="860" alt="screen shot 2017-02-09 at 18 27 04" src="https://cloud.githubusercontent.com/assets/91365/22797372/de4ddc52-eef5-11e6-8d69-23a5427f46e1.png">

**2) User clicks on a results and sees a back button**
![back](https://cloud.githubusercontent.com/assets/91365/22797417/0bcb9368-eef6-11e6-9359-1a98a19123f4.jpg)

**3) A user arriving to the object page **via any other means** doesn't see the back button (_maybe they see a home button instead, that just goes to the home page - TBD?_)**
![home](https://cloud.githubusercontent.com/assets/91365/22797437/1e09fe7a-eef6-11e6-94fd-3b7f8c3abcb8.jpg)

tobystokes:
One approach might be 'the crumb'
![image](https://user-images.githubusercontent.com/109024/27428195-3c456a5e-5739-11e7-9f87-5b2f9e007c3c.png)

I'm not sure whether the 'Home' link is useful though, and without an accompanying 'back' link, may look a little lost. 

As back button:
![image](https://user-images.githubusercontent.com/109024/27428124-0379841c-5739-11e7-93db-ec3bba788a84.png)
Could expand to 
![image](https://user-images.githubusercontent.com/109024/27428141-0ec86d60-5739-11e7-95a4-da83f7b6a9df.png)
on hover
(trying to visually associate it with the filters, though that may not be obvious)

jamieu:
[...]

PR #898, opened 2017-06-30
Back button

Danwhy:
## To merge after #896 
ref #722 
Adds a button to resource pages that leads back to the search page with the query performed if the page was accessed via search, or to the homepage if the user came directly to that page from outside the site.

Requires icons/styling from @tobystokes


tobystokes:
Thanks @Danwhy 
I suggest not merging this until I have added design into this branch too

client/routes/resources.js
```diff
@@ -47,12 +48,12 @@ function load (ctx, next, type) {
...
+      ctx.state.data.back = sessionStorage.getItem('backPath');
       // analytics
       window.dataLayer.push(JSON.parse(data.layer));
       window.dataLayer.push({'Level5': data.title});
 
-      ctx.state.data.page = type;
+      ctx.state.data.page = page;
       next();
     });
   } else {
```
shouston3: Do we need to change `type -> page` inside this `else` statement too?

Danwhy: :+1: Good spot

PR #900, opened 2017-06-30
Removing gitignored files

shouston3:
These get updated every time the tests run and cause our branch become dirty even though these files are `gitignore`d


SimonLab:
The tests are failing because Travis can't find the fixtures files anymore. @shouston3 we might need to find a better way to ignore the fixtures

Issue #779, opened 2017-02-02
Description boxes for key result pages


jamieu:
This builds on the work around nice urls and titles for key result pages (https://github.com/TheScienceMuseum/collectionsonline/issues/668)

Obviously we need some design (below is simply a rough mockup of idea) and some wider thought around what goes within that block.

But the basic idea is that for each Category, Museum and Gallery SERP 'landing' page we allow a short chunk of rich test to be displayed at the top of the first page or results. 

Initially this will just be the title and a description paragraph and some links, but would be good to allow room for linking to related articles (within the group sites) and/or show sub-catergories for large categories based on date or materials (ie. pre-internet / post-internet for communications or metal vs plastic medical instruments).

Need to decide what to show if we are filtering by both a category and gallery (which description block do we show (for now lets go with the category taking priority). 

This will also help with SEO as it creates nice results summaries for each category, museum and gallery within google.

The Thumbnail and Description fields (where available) should also be passed to the existing/relevant Twitter/OpenGraph/Schema.org metadata fields (I think we already pass the machine generated page title which we should keep as is) --> see this old ticket: https://github.com/TheScienceMuseum/collectionsonline/issues/570. 

![category-boxes](https://cloud.githubusercontent.com/assets/91365/22547898/db6b6b7c-e93b-11e6-9df9-7929dfb09b19.jpg)

![galley-summary](https://cloud.githubusercontent.com/assets/91365/24252756/4dcf72ec-0fd6-11e7-8b32-abf882347e46.png)

iteles:
#699 and #698 lead up to _this_ issue.

tobystokes:
did you have a spec doc for this, or was it my imagination?
The pics above will service, if that indicates all possible contents. Would be good to know what might be optional

jamieu:
The above will suffice for now (in giving an idea of content).

But @emilyfildes and I have had various conversations re. **exactly** what it should include, will have a read though my notes and update this issue.

jamieu:
Here goes, feel free to make suggestions!

But these are the ones we know we can produce content for / would like feature.
We won't have content for every field in every case; have marked these fields as _optional_

## Category 

- Title  
- Short description (1 or 2 lines max)
- Sub categories / queries (ie. all object of type 'transistor radio' in the 'telecommunications' category) [_optional / up to 4 or 5?]
- Related articles on main sites (_optional_) [_optional / up to 4 or 5?]

## Gallery 

- Title  
[...]

PR #902, opened 2017-06-30
Description boxes

shouston3:
- [x] Add `json` structure for description boxes
- [x] Pull the data through to render on the search page when filters are applied

See #779


shouston3:
Coverage has been increased in #910 instead

shouston3:
Just need to fix the tests then this should be good to go

jamieu:
This might just be me but when I go direct to the page (via a bookmark) I don't see the boxes? 

Although if I de-select and reselect the filter (or another one) on an already loaded search results page I do see it? Initial load vs AJAX?

- http://localhost:8000/search/museum/science%20museum/gallery/information%20age%20gallery:%20broadcast
- http://localhost:8000/search/museum/science%20museum/gallery/making%20the%20modern%20world%20gallery

client/lib/listeners/close-description-box.js
```diff
@@ -0,0 +1,12 @@
+module.exports = function () {
+  function select (elem) { return document.querySelector(elem); }
```
SimonLab: maybe the select function should be define outside the exported function

Issue #852, opened 2017-03-29
Revisit GA DataLayer


jamieu:
We don't seem to be getting any DataLayers sent back to GA, which we need to investigate urgently.

But, as part of this investigation it might be a good time to re-visit what we really want to track and ensure it is sent though in a nice/clean fashion.

## Object/Document/People pages
_(these are effectively singe record pages)_
- Title 
- Category (for objects)
- Name of containing Archive ie. 'Babbage papers' (for documents)

## SERP page
_(each modified 'free text' search query filter checkbox checked/unchecked or tab selected on this page should triggers a new GA event with a new DataLayer as if it was a new page request)_
- Search query (what appears in the search box)
- Type tab value (all / objects / documents / people)
- Image filter [filter value]
- Category  [filter value]
- OnDisplay [filter value]
- Object [type tab]
- Maker
- Place
- Archive
- DateTo
- DateFrom

_NB: We don't want to track pagination but should fire a new GA event each time someone clicks though to a new results page via the pagination._

jamieu:
Notice we still have `window.dataLayer.push({'Level5': data.title});` in https://github.com/TheScienceMuseum/collectionsonline/commit/a5027325c340a0066937bfbf53687030f2f2228c#diff-8c94a015faa4ac26fef59378f3bc0e1e 

Not sure we'll ever need need to use `Level5` etc. and think it might just be confusing us all. I think capturing there url and sending any other fields in a one level deep `data layer` is probably sufficient.

Also did you add a `param` to let us test in `staging`? (looks like you might have is so what the param/value?)

Thanks

Danwhy:
I've added the new variables and data layer events to Google Tag Manager, so when the PR is merged, they should be ready to publish. 

@jamie if you go on to tag manager, I think you should be able to see the unpublished changes I've made (there should be about 27 of them), and you/someone else at the Science Museum can publish them. You can also use the preview to check they are working on staging.

As I was saying on the call, I think the best way to stop the views from staging coming through to the analytics is to add a filter to the 'filtered view' on Google Analytics that only includes the live site. That way, we can still look at the unfiltered view, which will work with staging.

Let me know if you have any other questions, or can't see the changes on Tag Manager.

Danwhy:
@shouston3

jamieu:
@Danwhy we've now filtered the staging-server domain/url from GA, so it should only show in the 'unfiltered view' of the data.

[...]

PR #915, opened 2017-07-05
ga datalayer

Danwhy:
ref #852
Adds relevant datalayer variables for search/record pages

### DIFF
--- a/lib/transforms/serp-data-layer.js
+++ b/lib/transforms/serp-data-layer.js
@@ -1,0 +1,34 @@
+module.exports = function (data) {
+  var result = {};
+
+  addToLayer('query', data.q);
+  addToLayer('type', data.type);
+  addToLayer('category', data.categoryFilters);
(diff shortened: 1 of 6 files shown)


lib/transforms/serp-data-layer.js
```diff
@@ -0,0 +1,34 @@
...
+
+  addToLayer('objectType', findSelected(data.selectedFilters.object_type));
+  addToLayer('maker', findSelected(data.selectedFilters.makers));
+  addToLayer('place', findSelected(data.selectedFilters.places));
+  addToLayer('dateFrom', findSelected(data.selectedFilters['date[from]']));
+  addToLayer('dateTo', findSelected(data.selectedFilters['date[to]']));
+
+  return JSON.stringify(result);
+
+  function addToLayer (key, val) {
```
SimonLab: I think I would define this function outside the scope but I understand this implies adding the result object as a parameter. I think it would be a bit more readable.

PR #918, opened 2017-07-06
Better prev pagination

shouston3:
See #916 

- [x] Fix `pagination.js` documentation
- [x] Add better numbers to previously go to

### DIFF
--- a/lib/pagination.js
+++ b/lib/pagination.js
@@ -2,7 +2,7 @@
 Create a list of pagination pages. Returns an array of pages, possibly with
 a `null` break item:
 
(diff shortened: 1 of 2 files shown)


lib/pagination.js
```diff
@@ -49,6 +67,8 @@ function createPage (pageNumber, currentPageNumber, createPageLink) {
   return page;
 }
 
+// range(0, 2) => [0, 1]
+// range(4, 7) => [4, 5, 6]
```
SimonLab: :+1:

Issue #875, opened 2017-06-08
Update tests to reflect new index and brand-updates branch


jamieu:
Due to the number of new records in the upcoming/updated index we may need to update some tests - in particular the two Selenium tests below.

```
 TEST FAILURE:  2 assertions failed, 104 passed. (3m 44s)

 ✖ disactive-facet.clienttest

   - Disactive Facet (6.787s)
   Timed out while waiting for element <.filter__box[value=photographer]> to be present for 3000 milliseconds.  - expected "visible" but got: "not found"
       at Object.module.exports.Disactive Facet (/Users/jamie/projects/collectionsonline/test/client/disactive-facet.clienttest.js:9:8)

 ✖ sticky-filters.clienttest

   - Sticky filters persist after new search (10.042s)
   Testing if the URL equals "http://localhost:8000/search/museum/National%20Railway%20Museum/gallery/Great%20Hall?q=electric".  - expected "http://localhost:8000/search/museum/National%20Railway%20Museum/gallery/Great%20Hall?
```q=electric" but got: "http://localhost:8000/search?q=electric"
       at Object.module.exports.Sticky filters persist after new search (/Users/jamie/projects/collectionsonline/test/client/sticky-filters.clienttest.js:14:15)
```

shouston3:
@Danwhy I'm having issues running these locally, after installing the kit, I'm arriving at this page: https://www.java.com/en/download/faq/chrome.xml

Saying that my chrome version no longer supports this 😢 

I don't fancy fixing these from travis outputs, maybe we can pair on this tomorrow or something?

PR #922, opened 2017-07-07
Fix client tests

shouston3:
Ref #875


test/client/end-to-end.clienttest.js
```diff
@@ -10,9 +9,9 @@ module.exports = {
       .assert.urlEquals('http://localhost:8000/search?q=ada')
       .click('.resultcard')
       .pause(1000)
-      .assert.urlEquals('http://localhost:8000/people/cp38764/ada-lovelace')
+      .assert.urlEquals('http://localhost:8000/objects/co8228143/fort-ada-alexandria-photograph')
```
SimonLab: I'm not sure if this test should be changed or if we should check the search boost to know why Ada is not the first result. @Danwhy any thought or do you think this result is normal?

Danwhy: Yeah, I think you're right that the test should be checking that a result we'd expect to be the top result is, and if not, we'd need to tweak the search parameters. I don't think we should do this until after the upgrade to es5 though, as the changes to the search will likely be outdated.

Issue #882, opened 2017-06-15
Category filter not working


jamieu:
Selecting "Plastics and Modern Materials" results in no-resulst, despite first results (on wider result set matching).
http://collection.sciencemuseum.org.uk/search/object_type/coat-of-arms?page[size]=50&page[type]=search

<img width="1129" alt="screen shot 2017-06-15 at 11 31 55" src="https://user-images.githubusercontent.com/91365/27177465-667780c6-51be-11e7-9c87-cc93ab38ac0e.png">

Cleop:
2 hours time estimate to investigate (capped at whatever progress is made by this point)

Danwhy:
Figured out the issue with this.

Because the categories don't have a _lower field in the index, to get them to work as lowercase queries, we're just re-capitalising the first letter of every word on the back end. The problem with this is when a word is not capitalised, which is rare, but the case here, with 'and'.

Hopefully, all instances of 'and' in the categories is lower case, and we can just exclude that when we uppercase the filter, but the best action would be to add a _lower analyzed filed to the index.

PR #923, opened 2017-07-07
does not uppercase "and" in categories

Danwhy:
ref #882
Keeps "and" lowercase in category filters.

### DIFF
--- a/lib/helpers/utils.js
+++ b/lib/helpers/utils.js
@@ -1,6 +1,14 @@
 module.exports = {
   uppercaseFirstChar: function (str) {
+    var exclude = ['and'];
+
     if (!str) return false;
-    return decodeURIComponent(str).split(' ').map(e => e[0].toUpperCase() + e.substring(1)).join(' ');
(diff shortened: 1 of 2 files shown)


lib/helpers/utils.js
```diff
@@ -1,6 +1,14 @@
 module.exports = {
   uppercaseFirstChar: function (str) {
+    var exclude = ['and'];
+
     if (!str) return false;
-    return decodeURIComponent(str).split(' ').map(e => e[0].toUpperCase() + e.substring(1)).join(' ');
+    return decodeURIComponent(str).split(' ').map(e => {
```
shouston3: The `e` argument is a little confusing, maybe `word` would be better

lib/helpers/utils.js
```diff
@@ -1,6 +1,14 @@
 module.exports = {
   uppercaseFirstChar: function (str) {
+    var exclude = ['and'];
```
shouston3: I like how this is easily extendable, should we add `or`, `with` and a few others to this list

Issue #913, opened 2017-07-04
Page Titles (for Categories/Galleries) are all defaulting to base text


jamieu:
We seem to have lost the nice titles that used Category / Gallery / Museum (see  https://github.com/TheScienceMuseum/collectionsonline/issues/699)

For example this should display `Telecommunications | Science Museum Group Collection` but instead we are just displaying the base Title.
http://localhost:8000/search/categories/telecommunications

This is critical (along with the nice rest'style URLs, description boxes and ensuring the SERP page are in the sitemap.xml file) for the pages category/gallery/museum search results pages being nicely indexed by Google). See previously closed issue here https://github.com/TheScienceMuseum/collectionsonline/issues/699

<img width="876" alt="screen shot 2017-07-04 at 17 16 30" src="https://user-images.githubusercontent.com/91365/27837500-8b9a0ff4-60dc-11e7-8e87-ba406302b1af.png">

jamieu:
LGTM 👍

PR #926, opened 2017-07-07
makes search title pages correct

Danwhy:
ref #913
Title pages for key searches now show correctly


lib/helpers/search-results-to-template-data/title-page.js
```diff
@@ -14,9 +14,11 @@ var nameFromFilter = function (selectedFilters) {
   var museum = selectedFilters.museum && Object.keys(selectedFilters.museum)[0];
   var gallery = selectedFilters.gallery && Object.keys(selectedFilters.gallery)[0];
 
+  var exclude = ['categories', 'museum', 'gallery', 'type', 'page[size]', 'page[type]'];
+
   // check that only the filter categories, museum or gallery are selected
```
shouston3: Is this comment out of date?

Danwhy: No, not really. The functionality hasn't changed, it was just a bug that was picking up default seetings such as page type and size

Issue #883, opened 2017-06-15
No results page should suppress all/people/object/document tabs


jamieu:
http://collection.sciencemuseum.org.uk/search?q=zzzzzzzzzzzz

<img width="458" alt="screen shot 2017-06-15 at 11 34 26" src="https://user-images.githubusercontent.com/91365/27177581-c6ffba62-51be-11e7-9754-92d48798179c.png">

Cleop:
When there are 0 results:
- [ ] Remove People/Objects/Docs tabs
When there are 0 results and 0 filters applied:
- [ ] Remove filter search/ clear filters section on the right

jamieu:
Much better 🥇

PR #928, opened 2017-07-07
hides tabs on no results page

Danwhy:
ref #883


templates/partials/global/search-main.html
```diff
@@ -6,9 +6,11 @@
       <h1 class="searchbar__strap">Search our collection</h1>
       {{> global/searchbox filters="filters"}}
 
-      <nav class="searchtabs" id="searchnav">
-        {{> global/search-nav}}
-      </nav>
+      {{#ifpage 'noresults'}}{{else}}
```
SimonLab: maybe a ```unless``` helper would be a bit better here.

Issue #940, opened 2017-07-12
3rd box missing on search results (Safari)


jamieu:
View any result page on staging using the new flex boxes.

<img width="1244" alt="screen shot 2017-07-12 at 16 50 53" src="https://user-images.githubusercontent.com/91365/28126754-54084fdc-6722-11e7-9d89-4cfe19f699f0.png">

<img width="1204" alt="screen shot 2017-07-12 at 16 49 00" src="https://user-images.githubusercontent.com/91365/28126755-540cb784-6722-11e7-9968-9064c53de51f.png">

jamieu:
Hi @tobystokes it fixed the third box but for some reason one further down the page (Moneywise) has shrunk.
http://collectionsonline-staging.eu-west-1.elasticbeanstalk.com/search

<img width="1089" alt="screen shot 2017-07-13 at 16 15 02" src="https://user-images.githubusercontent.com/91365/28173525-96c68c20-67e6-11e7-892a-c260b7fcc88d.png">

jamieu:
Have merged your PR but this still seems to be an issue on Safari?

<img width="954" alt="screen shot 2017-07-14 at 16 31 49" src="https://user-images.githubusercontent.com/91365/28218921-05cc65a6-68b2-11e7-9169-15c6148eb592.png">

jamieu:
Not seeing this now, so hopefully just a blip.

Closing

PR #946, opened 2017-07-13
further layout consequences of #940

tobystokes:


tobystokes:
If no further browser issues reported in the meantime (I can't see any obvious) then yes, please do

Issue #807, opened 2017-02-23
Couple of outstanding Zoom/OSD UI bugs


jamieu:
- Still seeing the zoom controls after exiting full screen / OSD mode on mobile (iOS)
- The black background doesn't cover the full screen in full screen / OSD mode (iOS)

![img_3506](https://cloud.githubusercontent.com/assets/91365/23256737/b2ba9db4-f9b8-11e6-8b9e-65bc16f14bc9.PNG)
![img_3507](https://cloud.githubusercontent.com/assets/91365/23256738/b2d44854-f9b8-11e6-8732-cb1f0f4e9fe8.PNG)

Cleop:
2hour time capped estimate for investigation

shouston3:
It seems that openseadragon [doesn't play well with ios and is quite buggy](https://github.com/openseadragon/openseadragon/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20label%3Abug%20ios)

['pre-full-page' and 'full-page' fire on desktop browsers and other device browsers but not on ios](https://github.com/TheScienceMuseum/collectionsonline/commit/e8d4a53e1c73eba7574f0284692c648abbff2acd#diff-d90d166a94c061094a974de2b884081dR49) which gives us nothing to react to in terms of running javascript when going to full screen mode.

It seems that osd somehow hands over the full screen functionality to ios browser which we don't have control over.

### The black background doesn't cover the full screen in full screen / OSD mode (iOS)

I noticed that with the [toolbar](https://openseadragon.github.io/examples/ui-toolbar/) had `position: relative` and the gap at the bottom of the screen was the same size as the the toolbar.

Adding `position: absolute` to this element got rid of this gap, but this messed with osd's tracking of click events and the buttons didn't work. Having a toolbar with position relative at the top of the full screen causes a gap at the bottom of the view, but it is needed for osd to ensure the correct positioning of the button click events.

### Still seeing the zoom controls after exiting full screen / OSD mode on mobile (iOS)

The toolbar are being correctly created when clicking to go full screen, but as toggling of full screen after going full screen does not fire any events our logic which removes these elements also can't run.

We could look to raise an issue on `osd` about this as I couldn't find one for the `full-page` events not firing (although I found quite a few similar issues).

jamieu:
Thanks @shouston3 

The background (on iOS) isn't a major an issue, so let's park that one for now and I'll keep an eye on the updates to OSD.

[...]

PR #947, opened 2017-07-14
Ios image bugs

shouston3:
Only apply safari ios fix when on ios device

Ref #807 

For some reason, safari desktop could crash when making an image full screen. The change should no longer effect desktop and will only apply to mobile


shouston3:
We've just realised there is another change which we implemented [here](https://github.com/TheScienceMuseum/collectionsonline/pull/938/files#diff-6fe3d60018c7fc45b36ebcccf50a7ebdR8) which is actually the culprit for the safari desktop issue

shouston3:
This issue should be fixed now and working on safari as before (not crashing)

jamieu:
Excellent 👍 Well done and thanks again for debugging that!

PR #963, opened 2017-07-20
add log to test if full page fires or not

jamieu:
Does this still need merging or is it an old branch?


shouston3:
Sorry this is old, I'll delete it

PR #1108, opened 2018-01-13
redirect to object when question searched

SimonLab:
When a search contains a question starting with "what", we are looking for objects answering the question and redirecting to this object.

The logic needs to be added on the handler and on the client side for it to work with/without javascript on the browser

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,5 @@ public/assets/icons/*
 test/fixtures/elastic-responses/database.json
 test/helpers/aggregations-all.json
 test/fixtures/elastic-responses/example-*
+
+package-lock.json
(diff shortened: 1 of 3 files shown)


jamieu:
Thanks so much @SimonLab !!

That all makes total sense now, noticed you also tidied up /routes/search.js - so thanks for that as well.

I owe you a beer, or two :-)

iteles:
@jamieu Did you mean to close this PR without merging it in? 😊

jamieu:
Oh no! Thanks's for spotting 👍 

I was using the mobile site and must have hit the wrong button, was wondering why it hadn't deployed to staging,

Issue #1194, opened 2018-10-15
Homepage issues (unwanted client side rendering?)


jamieu:
These all seem to be the result of the home page and about views being loaded client side rather than just being served server side (there is no need for them to be updated on the client, as there are no filters on either page).

The most clear example of this is running a search and then going back to the home page where the original search URL is still showing in the browsers URL bar. 

All three core browser behave slightly differently, Safari seems to be best but is still glitchy, Chrome worse and in IE11 we seem to have lost any client side rendering!

1) About Us link (not working occasionally) - **AJAX loading / routing?**

The About Us link sometimes fails to load, assume this is related to there above routing/client side rendering? This page can simple be loaded via a server side route.

<img width="562" alt="screen shot 2018-10-15 at 10 55 18" src="https://user-images.githubusercontent.com/91365/46944106-1f4bb980-d069-11e8-9c64-715f3c73ce6d.png">

2) Highlights block wrapping when returning to home page (ie. go to a search and hit back) - **AJAX loading / routing?**

<img width="448" alt="screen shot 2018-10-15 at 10 50 49" src="https://user-images.githubusercontent.com/91365/46944089-178c1500-d069-11e8-95fa-88ce3c29682a.png">

3) There's also the gradient bleeding though when you go back from a search, again could be result of AJAX loading (when we just want to display the view as a static page/route).

<img width="702" alt="screenshot 2018-11-01 at 11 14 09" src="https://user-images.githubusercontent.com/91365/47851121-90140500-ddcf-11e8-8cea-1976af8603ff.png">

tobystokes:
1, 2, & 4 are all part of the same issue: When changing between these routes, if the page is not an initial render, the new content is swapped in via ajax. But that means initialising functions or event listeners aren't being triggered. That's my synopsis, but I don't know enough about how the routing works to fix it. So yes it may need @Danwhy to point us in the right direction here.

SimonLab:
> About Us link (not working occasionally) - AJAX loading / routing?

Hello,

There is an error which I think break the routing to the about page
![image](https://user-images.githubusercontent.com/6057298/48082387-65b8b200-e1ea-11e8-9d34-28a149a0ce3b.png)

`data` is not defined as an object but we try to assign a value to a property:
https://github.com/TheScienceMuseum/collectionsonline/blob/7f6f743c4dd5e54ce2a715f5c9fb40a319f37d65/client/routes/about.js#L24-L26

[...]

PR #1208, opened 2018-11-06
Home and about pages

SimonLab:
ref: https://github.com/TheScienceMuseum/collectionsonline/issues/1194#issuecomment-436366415

server rendering the home and the about page instead of using "ajax"

@jamieu I think the PR is ok but I couldn't find the credentials of the database to test the search page on my localhost, just to make sure I haven't broken anything else.


jamieu:
Excellent 🥇 

This seems to heave resolved al the issues including the highlights wrapping, so going to merge in to approve and merge into master

Thanks Again @SimonLab 👍

Issue #1008, opened 2017-09-13
Dash (-) in category names causing no results to be returned


jamieu:
https://collection.sciencemuseum.org.uk/search/categories/penn-gaskell-collection

![unknown](https://user-images.githubusercontent.com/91365/30369695-45a37ee0-986d-11e7-9042-3478b3710939.png)

iteles:
May not be fixable but we need to dedicate some time to reviewing and finding/proposing solutions.

SimonLab:
We are replacing spaces by "-"  when building the url in the query parameter. This is to avoid having `%20`in the url (which represent spaces). This work well until the real name of the parameter contains a "-". On the backend we are converting back `-` to sapces to search with Elasticsearch.
 linked to https://github.com/TheScienceMuseum/collectionsonline/pull/951 and https://github.com/TheScienceMuseum/collectionsonline/issues/515#issuecomment-314836894

The issue here is Elasticsearch try to match the category `penn gaskell collection` instead of `penn-gaskell collection`.
> Category didn't match, ensure the exact category title was used

Looking for a way to match the category when the name contains `-`

SimonLab:
To allow testing via the website a category containing a "-" we need first to display more categories in the list of filter (by default elasticsearch return the 10 first aggregation). We can add a `size` value to the aggregation, for example: `terms: {size: 500, field: 'categories.name' }` in 
https://github.com/TheScienceMuseum/collectionsonline/blob/736a0f7c4dbd30100dc7bd71dc841833aaf73c05/lib/facets/aggs-all.js#L20

This will display all the existing category (there are less than 500):
![image](https://user-images.githubusercontent.com/6057298/54284474-e4a66880-4597-11e9-8531-fe073e5065a2.png)

The other categories containing a dash are:
- X-rays
- Medical Ceramic-ware
- Medical Glass-ware
- Pharmacy-ware
- Penn-Gaskell Collection

SimonLab:
We can first make sure that the category description is displayed. This information is taken from the json file: https://github.com/TheScienceMuseum/collectionsonline/blob/master/description-boxes/category.json
So we need to make sure the keys (containing dash) match the categories. For that we can match the key by replacing all the "-" by spaces. This can be done in 
https://github.com/TheScienceMuseum/collectionsonline/blob/736a0f7c4dbd30100dc7bd71dc841833aaf73c05/lib/create-description-box.js#L8-L23

by changing the function to:
```js
[...]

PR #1250, opened 2019-03-13
Category name with dash

SimonLab:
ref: https://github.com/TheScienceMuseum/collectionsonline/issues/1008#issuecomment-472481014

Make sure the filter on categories containing dashes work with Elasticsearch

### DIFF
--- a/lib/query-params/query-params.js
+++ b/lib/query-params/query-params.js
@@ -28,7 +28,9 @@ module.exports = (typeFormat, queryParams) => {
 
   // Objects
   Object.assign(result.filter.objects, queryParamsObjects(typeFormat, queryParams));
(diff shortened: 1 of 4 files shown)


lib/query-params/query-params.js
```diff
@@ -28,7 +28,9 @@ module.exports = (typeFormat, queryParams) => {
 
   // Objects
   Object.assign(result.filter.objects, queryParamsObjects(typeFormat, queryParams));
-
+  if (result.filter.objects.categories) {
```
nelsonic: @SimonLab please consider adding more checks to avoid a `ReferenceError`:
```js
if (result && result.filter result.filter.objects && result.filter.objects.categories)
```
This will not create a big "tax" on perf but it will avoid a unrecoverable error.

Issue #1111, opened 2018-01-24
Update to HAPI v17


jamieu:

Danwhy:
@jamieu I've just about finished with the Hapi upgrade, but just wanted to check which version of node the live site is running on, as hapi 17+ requires at least node v8.12.0

jamieu:
@Danwhy I just upgraded the staging server to 8.15.0 and it seems to be running fine.

Let me know if you think we should make the jump to node 10 or leave it at 8 for now.

PR #1251, opened 2019-03-13
Hapi  v17/18 update

Danwhy:
ref #1111 

Upgrades hapi to v18:
 - upgrades all necessary hapi dependencies
 - switches reply for response toolkit (`h`)
 - replaces all handler callbacks with async/await


.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: node_js
 
 node_js:
-  - '7.10.1'
+  - '8.12.0'
```
nelsonic: Consider bumping this up to LTS https://nodejs.org (v10) ? 🤔

auth/authentication.js
```diff
@@ -1,19 +1,17 @@
...
-  server.auth.strategy('jwt', 'jwt', false,
-    {
-      key: config.JWT_SECRET,
-      validateFunc: validateToken,
-      verifyOptions: { ignoreExpiration: true }
+exports.plugin = {
+  name: 'Authentication',
+  register: async function (server, options) {
+    server.state('token', {
+      isSecure: false
```
nelsonic: consider making this an environment variable with a default to `false` so that it can be configured in AWS without a code release. (_it's OK for now_)

auth/session.js
```diff
@@ -7,13 +7,17 @@ module.exports = () => ({
...
     auth: false
   },
-  handler: (request, reply) => {
+  handler: (request, h) => {
     if (validate(request.payload.username, request.payload.password)) {
-      const session = {valid: true};
+      const session = { valid: true };
       const jwt = JWT.sign(session, config.JWT_SECRET);
-      return reply.redirect('/').state('token', jwt, {ttl: 5 * 3600 * 1000});
+      console.log('GOOD');
```
nelsonic: please consider removing these `console.log('GOOD')` and `console.log('BAD')` or making them more specific so the person reading the logs can stack-trace/debug them. 😉

lib/get-full-archive.js
```diff
@@ -21,11 +21,10 @@ module.exports = (elastic, id, next, count) => {
...
-      return next(err);
-    } else {
-      return next(null, result.hits.hits);
-    }
-  });
+  try {
+    const result = await elastic.search(searchOpts);
+    return result.hits.hits;
+  } catch (err) {
+    throw err;
```
nelsonic: what error handler is "catching" this `throw err` ...?

package.json
```diff
@@ -10,24 +10,25 @@
     "async": "^2.6.1",
     "awesomplete": "^1.1.1",
     "aws-sdk": "^2.350.0",
-    "boom": "^6.0.0",
-    "catbox": "^8.0.1",
-    "catbox-redis": "^3.1.1",
+    "axios": "^0.18.0",
```
nelsonic: why axios? https://www.npmjs.com/package/axios ? 💭

Issue #1117, opened 2018-02-12
Stop HTML/Category specific logic imposing on API calls


jamieu:
https://github.com/TheScienceMuseum/collectionsonline/blob/master/routes/search.js

>              // match categories
>               if (value.query.q && (!value.categories['filter[categories]'])) {
>                 var q = value.query.q.toLowerCase();
>                 var qMatch;
> 
>                 if (keyCategories.some(el => {
>                   if (el.category === q || el.synonyms.indexOf(q) > -1) {
>                     qMatch = el.category;
>                     return true;
>                   } else {
>                     return false;
>                   }
>                 })) {
>                   return reply.redirect(request.path + '/categories/' + qMatch);
>                 }
>               }

iteles:
Linked to #1115 .

PR #1256, opened 2019-03-20
stops redirecting to category on json calls

Danwhy:
ref #1117


nelsonic:
Merging as @SimonLab is off for planned annual leave. 👍

Issue #1219, opened 2018-12-05
Random object page


jamieu:
1. Migrate to this repo so that it works for any route ie.

http://collection.sciencemuseum.org.uk/search/categories/materia-medica-&-pharmacology/slideshow 
http://collection.sciencemuseum.org.uk/random

_(should default to only showing records with images)_

2. Modify CS JS code to cache 100 records/images in browser on first API call

## Acceptance Criteria
+ [x] Migrate code so it works within the current app
+ [x] Allow addition of `/slideshow` to any category endpoint
+ [x] Kill existing route
+ [x] Pull in 100 records/images in first API call
~+ [ ] Caching~

Danwhy:
This can be tested by going to:
`/search/categories/materia-medica-&-pharmacology/slideshow`
(or any category)
or
`/search/slideshow`

iteles:
Agreed with @jamieu that  as only 100 are loaded at a time, caching was no longer required.

PR #1263, opened 2019-03-29
removes random and screensaver routes

Danwhy:
ref #1219


Danwhy:
@nelsonic @SimonLab The coverage has dropped in this PR, but as I'm just deleting code, I'm not sure there's anything that can be done about that.

PR #1401, opened 2020-07-02
update the WDS to be compatible with plyr > 3.6

tobystokes:
small pkg bump to fix background colour on audio player


tobystokes:
👀  @jamieu

PR #1423, opened 2021-03-10
A11y

tobystokes:
(Apart from that initial update of Babel packages, which is worth worth checking still works on other environments) This should be mostly harmless HTML validation fixes.
This is sadly unlikely to be all accessibility work required, but it should get rid of the repetitive errors in the scans and allow us to see the remaining issues a bit more clearly.


tobystokes:
cc @argenteum

PR #1563, opened 2024-02-08
Related items images 1537

evancp87:
- fixed bugs with clicking through to related items and related items images not showing, the latter to do with a child records bug, which i also fixed
- fixed search weighting values for related items


jamieu:
Looks good 🎯

PR #1564, opened 2024-02-12
Child records design

tobystokes:


templates/partials/records/record-sph-item.html
```diff
@@ -0,0 +1,61 @@
...
+          {{{ this.value }}}
+          {{#if link}}</a>{{/if}}
+        {{/each}}
+      </dd>
+    </dl>
+    {{/each}}
+    {{# if parentTitle}}
+    <dl class="sph-records__dl">
+      <dt>Part of:</dt>
+      <dd>{{ parentTitle }}</dd>
```
tobystokes, line 52: As these cards have an id=[object number], it would be _nice_ if these `parentTitle`s linked to them - but because this partial only receives the 'current' child, not easy to get in handlebars. But if it was easy to pass in as `parentID` or something from the API...

templates/partials/records/record-sph-item.html
```diff
@@ -0,0 +1,61 @@
+<article class="sph-records__item" id="{{getinventorynumber details}}">
+  {{#if images}}
+  <a href="{{firstImage images 0 'large' }}" data-lightbox>
+    <figure class="sph-records__figure">
+      <img src="{{firstImage images 0 }}" class="sph-records__img" alt="{{this.title}}" />
+    </figure>
+  </a>
+  {{ else }}
+  <figure class="sph-records__figure">
```
tobystokes, line 9: previously, I only showed this fallback figure if any of the children had an image. Now this would need to be calculated from various levels of children, again might be better calculated pre-handlebars?

Issue #1634, opened 2024-03-07
CO2.0 : Missing child records on SPH [critical]


jamieu:
Only two child records are showing on this SPH record, but there are 168 Parts records in Mimsy.
http://localhost:8000/objects/co199371/genoese-medicine-chest-medicine-chest

[Currently the CIIM is limiting these to 25](https://github.com/TheScienceMuseum/collectionsonline/issues/1633) due to hard upper limit imposed by CIIM. But that doesn't explain why only two are showing currently.

<img width="791" alt="Screenshot 2024-03-07 at 12 36 37" src="https://github.com/TheScienceMuseum/collectionsonline/assets/91365/47c00517-c6df-45e9-8248-6489b004751d">
<img width="775" alt="Screenshot 2024-03-07 at 12 36 19" src="https://github.com/TheScienceMuseum/collectionsonline/assets/91365/0a81adbc-ac0e-4026-85b1-0f50c89f9a78">

evancp87:
@jamieu That should be ok, but remember I was having the issue where when i remove the count for MPH it wouldn't show any MPH records at all, so maybe should check if that's still happening?

jamieu:
> when i remove the count for MPH it wouldn't show any MPH records at all, so maybe should check if that's still happening?

Yes, we still need to hardcode a size=10,000 limit to that query, I just meant we don't (I think?) need two vary it depending on MPH vs SPH type.

 As the upper limit s now set in the CIIM / the SPM or MPH parent record, which will only even have XX record attached to it. Current 25 for both, but will likely get the SPH limit increased to 100 (or a nice multiple, post launch).

PR #1636, opened 2024-03-07
fixes only two record displayign issue on an SPH

jamieu:


jamieu:
Resolves https://github.com/TheScienceMuseum/collectionsonline/issues/1634

Issue #1656, opened 2024-03-12
Image Sort Order


jamieu:
Needs further investigation, 

This may be correct and we can close the ticket, but lets keep an eye on it for bit in case issues are spotted.

But I think we normally want to show the earlier modification (creation) date first, as the main image is precident uploaded first to the DAMMS. 

In 99.99999% in cases the `modification date` is the same as the `creation date` (for image assets) as we don't change the asset once uploaded. There may be some rare edge cases eg. in the past, prior to the `position` field, we sometimes 'fudged' the `modification date` to get the desired sort order. The `modification date` will mirror `creation date` in the DAMMS unless explicitly set (which  normal user can't do) so we can sort **solely** on `modification date`.

The cleaner should precede the attachments, although this may be an edge case
https://collection.sciencemuseumgroup.org.uk/objects/co8412068

<img width="1369" alt="1697092253000 - cleaner" src="https://github.com/TheScienceMuseum/collectionsonline/assets/91365/46770289-cfd5-4b6d-b4ab-91a7a764f33d">
<img width="1384" alt="1697092254000 - attachment" src="https://github.com/TheScienceMuseum/collectionsonline/assets/91365/ca3ee158-a4c0-4526-a2e8-39433b961aa5">

jamieu:
resolved, now uses `upload_sort` field on images

PR #1671, opened 2024-03-18
fix sort order

jamieu:


jamieu:
https://github.com/TheScienceMuseum/collectionsonline/issues/1656

PR #1698, opened 2024-04-08
update SMG map to include SIP

tobystokes:
Slightly shortened the description to match the other lengths, this and the link are now more easily editable in [fixtures/museums.js](https://github.com/TheScienceMuseum/collectionsonline/compare/sip-map?expand=1#diff-7f272c8d58533a9657a7d824fcc87de9856cd5f592a6da583e04b0358cf8d93e)


jamieu:
Thanks Toby, you're a 🌟

PR #1786, opened 2024-06-27
Design updates for groups and parts

tobystokes:
(in progress)


lib/transforms/search-results-to-template-data.js
```diff
@@ -360,6 +360,7 @@ function getCollageData (item) {
   } else {
     images = null;
   }
-
+  // creates full URL, like the figures
+  images = images.map((image) => mediaPath + image);
```
tobystokes, line 364: @evancp87 These paths will need to be the full URL to the media server. Couldn't work out how/where the  other images get this appended! My hack here works for the initial load but not the client-side rendering

evancp87: So for top level record images it seems the full path to the media server is set in search-results-to-jsonapi.js, so i think i will need to do the same for child records also.

PR #1791, opened 2024-07-03
feat: configured full path to image server

evancp87:
@tobystokes The full path to the images should now be built correctly, at the api level, in line with how it is done on record pages


tobystokes (approved):
Yes it does! Ta.

PR #1797, opened 2024-07-26
explore/group banners, etc

tobystokes:


templates/partials/records/record-top.html
```diff
@@ -1,3 +1,15 @@
+{{! this test needs to be updated for the url of a banner image. If those images are added as fixtures, ideally include whether the image is predominantly dark or light, so that the text can be styled accordingly. }}
```
tobystokes, line 1: still a bit of work to do here @jamieu , not sure where this image is going to come from...

PR #1801, opened 2024-08-05
fix pagination links

tobystokes:
fix wrong url introduced in 50b7af08


jamieu:
Thanks Toby, I think I just opened almost exactly the same PR ;-)
https://github.com/TheScienceMuseum/collectionsonline/pull/1802

So closing hits one, but thanks for responding so quickly..

PR #2196, opened 2026-08-11
Compact the Themes from the collection cards on phones

jamieu:
Below the single-column breakpoint (580px) each theme got a full-width 4:3 image, so the pictures dominated the page and pushed the category links below the fold.

This switches to a row on phones: **5.5rem thumbnail on the left, title, description and links on the right**. At 375px a theme card goes from roughly 440px tall to 206px, so three themes fit in a viewport where one did before.

Unchanged at ≥580px, where the existing 2- and 4-column grids take over.

CSS only — one addition to `components/_home.scss`, scoped to `.home-cards` inside `@include media('<small')`. No markup or data changes; the thumbnail keeps its 4:3 box via the existing `card-image` mixin, just in a narrower column.

## Two things you may want to weigh in on

**Hyphenation.** The narrower text column triggers the design system's existing `<=small` hyphenation far more often — you get "develop-ments" and "com-muter". The rule is pre-existing (`components/_resultcard.scss`), just newly visible now the column is ~237px instead of ~341px. One line to disable for this layout if you'd rather have a ragged edge.

**The description.** Kept it, since removing copy wasn't part of the brief, but dropping it on phones would tighten each card further and leaves the thumbnail-plus-links pairing you described.

## Testing
- `semistandard` clean; 941/941 unit tests pass
- Verified in-browser at **375px** (thumbnail left, info right, top-aligned, figure 88×66 at a measured 1.33 ratio) and at **768px / 1280px** (unchanged — info still stacked below the figure)


jamieu:
Parking this for now — the phone layout needs more thought, so it shouldn't sit in review half-decided.

Closing rather than merging. The branch `acrn/themes-mobile-layout` is left in place, so reopening this PR restores the change exactly as it is here.

For whoever picks it up (probably me, tomorrow), the measurements at 375px:

| Variant | Section height | Tap targets |
|---|---|---|
| Current on master (stacked, full-width images) | 1959px | 19px |
| This PR (row: 88px thumb, descriptions kept) | 1029px | 19px |
| Row + descriptions dropped + 44px taps | 1063px | 44px |
| Row, no thumbnails | 932px | 19px |
| Row, no thumbnails + 44px taps | 1177px | 44px |

The unresolved question isn't the layout, it's the category links: they're 19px rows on a 30px pitch, under the 44px minimum, and that's true on master today too. Fixing it costs ~245px unless something else gives. Dropping the descriptions was rejected because a theme has no page of its own, so the description is the only thing explaining what it covers.

